PR #73, opened 2024-01-05
CV Pipeline Integration

### LINKED ISSUE #30: Camera Connection Mock
## 🚀 Feature
For future camera implementation testing, we need to create a mock camera connection that fetches images from test folders. 

## Test
Make sure the images are fetched correctly.

### LINKED ISSUE #62: Set up orchestrator for CV pipeline
## 🚀 Feature
Set up a tick function for the search phase that will take photos and pass images into the CV pipeline. 

## Pitch

TODO

atar13:
Closes #62 
Closes #30 

## Changes

Set up a skeleton of the Computer Vision pipeline. This includes classes and function signatures for all modules of the pipeline. Now, the types of output data of each stage of the pipeline is well defined.

Also created a `SearchState` class that inherits from `MissionState`. In its tick function it starts up a new thread that captures a new image from the camera and passes it into the pipeline. I have not implemented any way to collect results from multiple threads, send the results to the GCS, or shut down threads that exceed a timeout. These should be done in a later PR.

This PR also includes a `MockCamera` implementation that implements `CameraInterface`. I just added it to easily be able to fetch dummy image data. Right now it just returns an image with a solid color but in the future it should return an image with valid competition targets.

## Testing

There's a small unit test to test an image crop helper function. `tests/unit/cv/utilities_test.cpp`

There's another unit test to test the mock camera `test/unit/camera/mock.cpp`

There's an integration test to allow someone to run the entire CV pipeline using an arbitrary image from a filepath. `tests/integration/pipeline.cpp`

## Feedback

Any feedback is appreciated, but I'd like to focus on a few specific parts. 
- I tried to implement some version of dependency injection with the `CameraInterface`. It can be found [here](https://github.com/tritonuas/obcpp/blob/feat/cv-orchestrator/include/core/states.hpp#L76) based on the first approach in [this blog post](https://vladris.com/blog/2016/07/06/dependency-injection-in-c.html). 
- I also would like some feedback on how I'm spawning the pipeline thread and the comments right above about thread safety https://github.com/tritonuas/obcpp/blob/feat/cv-orchestrator/src/core/states.cpp#L48-L81. 
- Also if there's a better way to represent the [results of the Pipeline](https://github.com/tritonuas/obcpp/blob/feat/cv-orchestrator/include/cv/pipeline.hpp#L28-L36), I'd like to hear it.
- Any other C++ idioms/standards I violated especially with any OOP stuff


Tyler-Lentz (approved):
I really like all of the code structure and organization for everything. I had the issue with the image paths, but I think that's just a me issue and I don't think its worth holding up the merge over it. Once we automate the asset pulling process fully we can rigorously test to make sure everything is robust.

PR #104, opened 2024-01-31
Add dynamic avoidance reaction

Samir-Rashid:
Greetings, gamers.
This PR desperately needs some CPR. I am requesting some backup in reviving this. As it is right now, it is just skeleton code. I think the comments express the idea I was going for. (TLDR: camera is always taking pics in another thread and this should trigger a state transition iff in airdrop state to emergency avoidance state which changes altitude until it cannot see an obstacle)

There are a few issues:

- [x] 1. This PR depends on @hashreds PR to get finished so that I can do inference with SAM
- [ ] 2. This PR depends on the camera code that @dzodkin33 wrote. There is going to be a refactor at some point, but that is not currently scheduled.
- [ ] 3. Also we don't even know what camera is going to be used yet! Is there even a spot on the plane? That camera needs to get code added to take photos.
- [ ] 4. I do not know how to test the custom mavlink command code I wrote. There is a special message to change altitude and continue along a waypoint path which I want to use. This is not in mavsdk, so I had to use the passthrough.
- [x] 5. we need to compile mavlink to use plugins (Passthrough is a plugin)
- [x] 6. I have to update to the newest Tick architecture (should not be hard)


include/core/evasion_state.hpp
```diff
@@ -0,0 +1,71 @@
...
+        // List of obstacle locations.
+        std::vector<GPSCoord> obstacles;
+        /*
+            Do the evasion logic and send mavlink commands.
+        */
+        void evade(void);
+};
+
+
+    // Wait until health is OK and vehicle is ready to arm
```
Samir-Rashid, line 60: Ignore this

include/core/evasion_state.hpp
```diff
@@ -0,0 +1,71 @@
...
+    will choose a form of active evasion and then return control to normal
+    pathing afterwards.
+    The path planning does not run in real-time, so avoidance behaves by inserting
+    emergency waypoints and then returning to continue the next unhit waypoint.
+*/
+class ObstacleEvasionState {
+    public:
+        // Do not use a default constructor. The evasion state needs to know
+        // the location of the obstacle
+        ObstacleEvasionState() = delete;
```
atar13, line 29: I don't know if you need to delete the default constructor. I think if you define another one it will not generate a default one. But deleting it shouldn't hurt.

include/cv/dynamic_detection.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+Threads:
+Thread 1: Runs steps 1-3.
+Thread 2: This is the main program execution, which should be polling every tick.
+*/
+
+// Output after an obstacle has been detected
+struct Detection {
+    std::chrono::time_point<std::chrono::system_clock> time;
+    // TODO(Samir): is there a type for location of bounding box
```
atar13, line 39: There is a bounding box but it stores two corners of the bounding box, not the center of the bounding box.

https://github.com/tritonuas/obcpp/blob/664bbc81eb86ff18b0b7a647ade1d08fb56769cf/include/cv/utilities.hpp#L20-L29

include/cv/dynamic_detection.hpp
```diff
@@ -0,0 +1,86 @@
...
+     // We need to make sure that a detection is not spurious
+     // before engaging in emergency evasion.
+     std::queue<Detection> queue;
+     // Note that Step 4 in "Steps" above only triggers on a boolean
+     // condition. I am still making this Detection queue thread safe
+     // because it can be useful to be able to access this in the main
+     // program. For example, once emergency evasion is engaged it can
+     // poll the queue to make sure there is nothing more to evade.
+     std::mutex mutex;
+};
```
atar13, line 65: I like hiding the complexity of the queue a lot and exposing a thread safe interface. It would be nice if we could have a generic version of this for all of obcpp since so many places are using thread safe queues.

src/cv/dynamic_detection.cpp
```diff
@@ -0,0 +1,89 @@
...
+    // Object coming closer, multiple consistent detections
+    // can do some housekeeping here to make sure queue
+    // does not have stale (>15s) detections
+    
+    // do housekeeping: remove stale detections
+    ObstacleDetectionQueue::pruneQueue();
+
+    // TODO: check if the boxes are getting larger AND bigger than a threshold
+    std::queue<Detection> queue;
+    ObstacleDetectionQueue::copyQueue(queue);
```
atar13, line 62: What does the `::` syntax here do? I might be trolling with C++ but I thought you needed to create an `ObstacleDetectionQueue` instance to call methods on it?

src/cv/dynamic_detection.cpp
```diff
@@ -0,0 +1,89 @@
...
+        // This is waiting to see how the model returns bboxes (bounding boxes)
+        return true;
+    }
+    
+
+    // TODO: I am rethinking this now and it does not make sense to 
+    // have dynamic obstacle detection be part of it's own state.
+    // The dynamic detection should be a part of a separate thread and
+    // send a message to the Airdrop state to switch to the dynamic avoidance
+    // state.
```
atar13, line 74: I agree with this sentiment. I think it should be running in another thread during Airdrop and Search states.

Samir-Rashid:
I will close this PR unless anyone objects. Ultimately, I do not believe vision is a satisfactory approach for dynamic avoidance. It would not be the worst to adapt the ideas from this PR to next year. However, we need to have a proper LiDAR solution at some point.

PR #126, opened 2024-04-01
Feature RRT*

### LINKED ISSUE #59: RRT* Algorithm Implementation
## 🚀 Feature

It is time to implement RRT*. I will be referring to the old python implementation throughout this issue, but I would strongly recommend restructuring to whatever architecture makes the most sense to you as you develop it, as long as it follows the high level interface I specify below. (And I would strongly recommend splitting into much more specialized functions than what is in the old code. Ideally the high level functions should basically read like pseudocode in that all the function names just say what they are doing.

## High level view of RRT algorithm

```
Input: [start: RRTPoint], [goal: RRTPoint], [num_iterations: Integer], [goal_percentage: Double]

start: starting point (with angle psi value)

goal: point we are trying to reach. psi value is unimportant, as we can enter from any angle. Perhaps we identify an "ideal" angle of approach depending on if point is close to the edge (so we don't trap ourselves into a corner).

num_iterations: number of points we randomly sample and attempt to connect to the tree before checking if we have reached the goal

goal_percentage: percentage chance that a sampled point is actually just [goal]

------------
Attempt to directly connect [start] to [goal] with dubins curve. 
The old implementation randomly samples 8 angles of approach into [goal], but we could do something smarter. 
If we get a path that meets [EARLY EXIT CONDITION: some threshold of distance that we call "good enough"
     (e.g. the path distance is no larger than 5% of the straight line distance)], then we return early.

Repeat [num iterations] times:
    Randomly sample a point (call it [sampled point]), with a [goal percentage] chance to sample [goal].
    Attempt to connect the point with the tree using dubin's algorithm. 
         Get a list of possible edges that connect [sampled point] with the RRT tree. 
         Going through the list in order from shortest to longest, iterate through connecting points and verify all are in bounds.
             (if you are reading the python code, note that an else statement following a for loop only runs if the for loop was not broken out of)
    If a path has been not been found, return to begin of loop and start new iteration.
    Connect [sampled node] into the RRT tree.
    Check if introducing this sampled node allows for any rewiring to optimize the tree (RRT*)

Return generated path (In RRTPoints)
```

RRT* can be summarized to this:

Assume we just added [sampled point] to the tree.

```
For [curr point] in RRTTree where [curr point] is within [NEIGHBOR_DISTANCE] to [sampled point]
      Generate possible dubins paths to connect [curr point] to [sampled point].
      Iterate through each dubin's path, checking for out-of-bounds, in order from shortest path to longest path.
      If the edge that connects [sampled point] to [curr point] is shorter than the edge that currently feeds into [curr point],
           then rewrire the tree such that [sampled point] replaces [curr point]'s old parent node.
```

## Levels of abstraction

The lowest level of abstraction is encapsulated by the old implementation's [rrt run method](https://github.com/tritonuas/obc/blob/main/pathing/rtpp/rrt.py). This is the pseudocode I described above. Note that in the old python implementation this was another method on the `RRT` class itself. We could do it this way (tacking it onto the RRTTree), or we could write a separate free standing function called `rrt` which completely encapsulates the `RRTTree` itself. (i.e. the user does not need to instantiate an `RRTTree`, they can just call the `rrt` method. Im leaning towards the latter because I think having the `RRTTree` solely deal with the operations of saving data in the tree makes it simpler.) You could also make a separate RRT class that composes an RRTTree, or any number of things. Do whatever you like / think makes the most sense / looks the nicest / what a random coin flip tells you.

The higher level of abstraction is encapsulated by the old implementation's [generate_generic_path](https://github.com/tritonuas/obc/blob/main/pathing/rtpp/paths.py) method. Essentially, at this level we have some sort of `rrt` function that takes two RRTPoints and returns a path connecting them. __The input to this function is a list of GPSCoords, which represent the list of competition waypoints. Michael is nearly done with his cartesian conversion code, so until that is done you can just make the input XYZCoords.__ The output will similarly be GPS coords. Essentially, this function should call the lower level rrt function between each pair of adjacent waypoints and stitch together the final path.

## Interface

By the end of this, the only function outside code will call will be our version of `generate_generic_path`, so that functions signature is important, but any lower level functions you need to add or modify won't be used by the rest of the code so any modifications there are totally fine.

The high level function should look like this:

```
std::vector<GPSCoords> generateInitialPath(std::shared_ptr<MissionState> state) {
     // do pathing here
}
```

AskewParity:
closes #59 

# Static Path Planning

This feature is the algorithm governing static path planning. _Inspired_ by RRT and RRT*, it uses random sampling to find a path to a goal .

## Algorithm

The algorithm by connecting waypoints one by one, having no knowledge of work done to connect previous waypoints. Between two waypoints, the algorithm goes

> All connections are valid Dubin's curves

1. Attempts to connect the two waypoints directly
2. If fail, it starts to populate a tree of paths with a number of $n$ nodes, where
	- it connects a uniformly sampled point within the valid area to the existing tree
	- RRT* is ran to rewire the tree to be optimal
	- 5 times, there is an epoch evaluation, where we attempt to connect to goal, if there is insufficient progress from one epoch to the next, we exit early
3. If we have not yet exited, we attempt to connect to goal with the full tree

## Options

`optimization` - whether to run RRT* or not (doesn't have a massive impact from experience?)

`point_fetch_method`
- `NONE` will check RRT against every node (path optimal, but incredibly slow)
- `RANDOM` will check ~$k$ randomly sampled nodes from the tree.
- `NEAREST` will check ~$p$ nodes closest to the sampled node (best performance/time ratio from rudimentary testing) 

`allowed_to_skip_waypoints`
- true --> will skip waypoints
	- e.g. if connecting from 1-2-3, fails to connect 1-2
	- Tree created from 1 will be retained
	- Attempts to connect 1-3
- false --> will keep running RRT until a connection can be found, will be much more thorough with checking for goal connections if the initial RRT iteration fails.

## Auxiliary Classes

Tree
- Governs the behavior of the tree in RRT and preforms RRT* by rewiring itself
- vastly different from the tree originally pushed to main

Environment
- Describes the information about the airspace (valid/invalid regions, waypoints, etc.)


tests/integration/path_planning.cpp
```diff
@@ -0,0 +1,338 @@
...
+#include "pathing/plotting.hpp"
+#include "pathing/static.hpp"
+#include "ticks/mission_prep.hpp"
+#include "ticks/mission_upload.hpp"
+#include "ticks/path_gen.hpp"
+#include "ticks/tick.hpp"
+#include "utilities/constants.hpp"
+#include "utilities/datatypes.hpp"
+#include "utilities/http.hpp"
+
```
atar13, line 23: Not that big of a deal, but I think it could be nice to have a comment explaining what this integration test does and what files it outputs for someone looking to test path planning in the future.

tests/integration/path_planning.cpp
```diff
@@ -0,0 +1,338 @@
...
+    std::cout << "Start Plotting" << std::endl;
+
+    // for (const XYZCoord& point : path) {
+    //     file << point.x << ", " << point.y << std::endl;
+    // }
+
+    PathingPlot plotter("pathing_output", state->config.getFlightBoundary(), obstacles[0], goals);
+
+    plotter.addFinalPolyline(path);
+    plotter.output("test_final_path", PathOutputType::BOTH);
```
atar13, line 331: Jeez, the plotting takes forever. We should really figure this out to make debugging easier. Made an issue finally so we can look into it. https://github.com/tritonuas/obcpp/issues/128

atar13 (approved):
Amazing work with seeing this through and really owning this project. This is a monumental part of our system and you did a great job putting it together from basically nothing.

tests/integration/path_planning.cpp
```diff
@@ -232,19 +231,32 @@ const static char* mission_json_2020 = R"(
   ]
 })";
 
+/*
```
Tyler-Lentz, line 234: I really like this diagram

Tyler-Lentz (commented):
This PR should be ready to merge at any time, so once you're confident that there's no other tinkering you want to do you should be able to merge it in, since it is approved. 

If you end up merging after the CV pipeline branch gets merged in you might have to repull from main, but I don't think that should lead to any major merge conflicts.

EDIT: actually we are just gonna merge it now and you can include any fixes later in the second pathing branch we talked about at the meeting.

PR #138, opened 2024-04-05
Refactor/pipeline results

Tyler-Lentz:
This PR refactors how the results of the pipeline are exposed throughout the rest of the system, especially from the matching part of it. Essentially we expose the actual similarity score and the pipeline just returns back the raw results, without making too much of a judgement beyond reporting whatever bottle the target matched with the closest.

This also will interop better with how the GCS is expecting the target information to be sent down.

The main inclusion of this PR is the CVAggregator class, which is a class that now lives inside of mission state and allows the other parts of the component to run the pipeline on a given image.


CMakeLists.txt
```diff
@@ -162,6 +162,23 @@ target_add_loguru(cv_pipeline)
...
+target_add_matplot(cv_aggregator)
+target_add_opencv(cv_aggregator)
+target_add_loguru(cv_aggregator)
+target_add_httplib(cv_aggregator)
+target_add_protobuf(cv_aggregator)
+target_add_torch(cv_aggregator)
+target_add_torchvision(cv_aggregator)
+target_add_mavsdk(cv_aggregator)
+target_add_loguru(cv_aggregator)
+# for some reason calling target_add_imagemagick here conflicts with, so we are including/linking without the function call
```
Samir-Rashid, line 177: average CMake feature

include/cv/aggregator.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <vector>
+#include <future>
+#include <queue>
+#include <array>
+#include <functional>
+
+struct CVResults {
+    std::vector<DetectedTarget> detected_targets;
+    // matches[0] = 5 => detected_targets[5] is matched with bottle A
+    std::array<size_t, NUM_AIRDROP_BOTTLES> matches;
```
Samir-Rashid, line 19: what is the point of using [`size_t`](https://en.cppreference.com/w/cpp/types/size_t)?

src/cv/aggregator.cpp
```diff
@@ -0,0 +1,95 @@
...
+    Lock lock(this->mut);
+
+    if (this->num_worker_threads > MAX_CV_PIPELINES) {
+        LOG_F(WARNING, "Tried to spawn more than %ld CVAggregator workers at one time.", MAX_CV_PIPELINES);
+        this->overflow_queue.push(image);
+        LOG_F(WARNING, "CVAggregator overflow queue at size %ld", this->overflow_queue.size());
+        return;
+    }
+
+    static int thread_num = 0;
```
Samir-Rashid, line 33: Reviewing C++ is so difficult. I would have never caught it if you got this `static` wrong, or any other locking or memory issue.

src/cv/aggregator.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+void CVAggregator::worker(ImageData image, int thread_num) {
+    loguru::set_thread_name(("cv worker " + std::to_string(thread_num)).c_str());
+    LOG_F(INFO, "New CVAggregator worker spawned.");
+
+    while (true) {
+        auto results = this->pipeline.run(image);
+        
+        Lock lock(this->mut);
```
Samir-Rashid, line 46: I do not understand the behavior of the CVAggregator mut. What is this protecting? I also am unsure how this mut behaves when you have threads.

tests/integration/cv_aggregator.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+// this image should be located at a relative path to the CMake build dir
+const std::string imageTestDir = "../tests/integration/images/matching_cropped/test/";
+const std::string refImagePath0 = imageTestDir + "000000910.jpg"; // bottle 4
+const std::string refImagePath1 = imageTestDir + "000000920.jpg"; // bottle 3
+const std::string refImagePath2 = imageTestDir + "000000003.jpg"; // bottle 2
+const std::string refImagePath3 = imageTestDir + "000000004.jpg"; // bottle 1
+const std::string refImagePath4 = imageTestDir + "000000005.jpg"; // bottle 0
+
+// matching model can be downloaded from here: https://drive.google.com/drive/folders/1ciDfycNyJiLvRhJhwQZoeKH7vgV6dGHJ?usp=drive_link
```
Samir-Rashid, line 20: uh..
> and we aren't exposing any of the real trade secrets (CV/ML models)

Samir-Rashid (approved):
I like the minor refactorings. I generally agree with the concerns about bottle similarity scores/thresholds. The logging additions are great.

include/cv/aggregator.hpp
Tyler-Lentz: It is more explicitly an index into a container, compared to an int which is very general.

src/cv/aggregator.cpp
Tyler-Lentz: Each worker thread is able to run the pipeline independently because they are just accessing the models and running the algorithms on the images independently, but once the results have been extracted from the pipeline the worker thread needs to put the results back in the two state variables in CVAggregator: the vector of all detected targets and the array containing match information. So this logic of going into the CVAggregator data members and modifying the state needs to be a critical section so that two workers don't try and modify the aggregated results at the same time.

tests/integration/cv_aggregator.cpp
Tyler-Lentz: uhhhhhhhhhhh

Tyler-Lentz: should we try and obfuscate this @atar13

src/cv/aggregator.cpp
Tyler-Lentz: Each worker thread is running the private function "worker", so they all have access to the shared memory contained within the `this` pointer. So they all can lock the same mutex and append to the same state variables inside of CVAggregator.

src/cv/aggregator.cpp
```diff
@@ -0,0 +1,96 @@
...
+                LOG_F(INFO, "Made first match between target %d and bottle %d",
+                    detected_target_index, curr_target.likely_bottle);
+
+                this->results->matches[
+                    static_cast<int>(curr_target.likely_bottle) - 1] = detected_target_index;
+            } else if (curr_match->get().match_distance > curr_target.match_distance) {
+                LOG_F(INFO,
+                    "Swapping match on bottle %d from target %ld -> %d (distance %f -> %f)",
+                    static_cast<int>(curr_match->get().likely_bottle),
+                    this->results->matches[static_cast<int>(curr_match->get().likely_bottle)],
```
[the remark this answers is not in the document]
Tyler-Lentz: incredibly good catch. holy. I think actually im just going to make it a hashmap instead of an array because the performance difference will be incredibly minor and it will be much more readable

PR #139, opened 2024-04-06
The Great CMake Library War

Tyler-Lentz:
> I am become CMake, builder of worlds.
>                 - Robert Oppenheimer, 1945

## Motivation

Currently, each individual make target completely recompiles all of the obcpp source files. This means that if you just do `make` then you are compiling the same files in the `src` directory numerous times, one per make target. This is wasteful in both time and space.

This PR reworks the CMake setup of the repo so that the obcpp related code is packaged as a shared library, so all of the testing targets can use the same `.so` files. This dramatically reduces redundant compiling.

## Changes

> I looked upon the world and thought, "Why are there only 2 CMakeLists.txt files? Why not 4? Or perhaps 11?"
>                   - Proverb (unknown, c. 5th century AD)

Each individual subdirectory in the `src` folder is packaged as a library (wherein `src/camera` becomes `obcpp_camera`, etc...), and all of these libraries are ultimately packaged under one `INTERFACE` library called `obcpp_lib`.

Each of these libraries explicitly defines the files that it includes. I think this is a lot more managable because there is one CMakeLists.txt file per directory, and if you ever add/rename/delete a file you only have to modify the CMakeLists.txt file in that directory. This also is more Nix-like, because the state of the repo is directory encoded in text files themselves, and not hidden in the `build` directory based on when you last ran `cmake ..`. I am just saying words. By getting closer to best Nix practices we are one step closer towards enlightenment (Rust).

## Tasks

- [ ] successfully compile/run the main `obcpp` binary
- [ ] successfully compile/run the unit tests
- [ ] successfully compile/run each of the integration tests
- [ ] rm -rf `build` and then run `make`, and have everything build/run correctly

## Potential Optimizations

Currently, I am linking every single dependency into every single shared library, to just make sure I can get it working. In the future, however, we could only link dependencies into a shared library if that library actually uses it.

## Anticipated Challenges

Before we were testing the network code by explicitly choosing which mock networks files to include and explicitly removing the files that depended on the network. Now, we will probably have to make alternative versions of certain libraries (namely `obcpp_network`) that have the mock implementations loaded in.

## Misc

I am also looking into [this cmake feature](https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html) to see if we can improve compile speeds. I will try to benchmark this once all of these changes have been made.


pink10000:
lgtm

atar13 (approved):
Thanks for looking into this speed up and keeping up the fight against cmake.

PR #142, opened 2024-04-07
Feat/pathing airdrop

### LINKED ISSUE #136: Airdrop Zone Pathing
## 🚀 Feature

Generates paths that "paint"s over the airdrop zone to search for targets.

AskewParity:
closes #136 

# Airdrop Zone Coverage Pathing

This code adds to `static pathing` a new class that's sole role is to make a coverage path over the airdrop zone.

There are a lot of problems enumerated below, but given the specifications of the competition, I think this naive method is ok to merge.

The code after implementing suggestions below would be over-engineered, and not really be portable to other applications (RRT in its current state is over-engineered, but its optimized form could be reasonably used for dynamic path planning).

## Problems

- shape must be a closed, convex, polygon to work correctly
- does not check whether the path is invalid
  - checking could be trivially added
  - alt could be to shrink the polygon by a factor of `turning radius` (would be fine if we are conservative with `search radius`)
- only does vertical from left-right
  - could be a config option
  - could test multiple options and take the most efficient

## 

- Not thoroughly unit tested 
  - the intersection code is unit tested, so as long as the `for` loops are written correctly, this should work
  - pushing under the guise of "trunk based development"


Tyler-Lentz (approved):
:+1: Looks very good, we should verify that this works well for the competition boundaries, but we can do that later on. Definitely want to merge this in asap.

The one commit I pushed was just because I did some changes to the mission upload tick and one of your includes in the integration test got broken, so I fixed that

PR #152, opened 2024-04-15
Feat/config json

### LINKED ISSUE #74: Config JSON
We need a `config.json` file which lives in the repo to keep track of parameters with which to run the OBC. Something in the code should load in the config file, then save the values somehow to be loaded into the rest of the program where needed. If the config file doesn't exist, a default one should be generated. In addition, we should be able to specify what config file to use on the command line by passing a command line argument, so that we can have separate config files for production (`config.json`) and development (`dev-config.json`). This is what the development config file should have as defaults:

```json
{
    "network": {
        "mavlink": {
            "connect": "tcp://172.17.0.1:5760",
        },
        "gcs": {
            "port": 5010,
        },
    }
}
```

And this is what the production `config.json` should look like:

```json
{
    "network": {
        "mavlink": {
            "connect": "serial:///dev/ttyACM0",
        },
        "gcs": {
            "port": 5010,
        },
    }
}
```

Step 1: try to load a config file based on the first command line argument

e.g. ./obcpp dev-config.json

Step 2: take the values from the config files, and load them into the appropriate parts of the OBC

 - GCS Port
 - Mavlink connection URL

Note: we are using this json library https://github.com/nlohmann/json

mobslayer202:
closes #74 

Made OBCConfig struct which loads configs from json file specified in command line argument

- throws error if json file not found
- uses default settings and generates default config json file when no command line argument used
- Struct passed into OBC constructor


src/main.cpp
AskewParity, line 1: hi

Tyler-Lentz:
I fixed some of the linting errors and made some styling changes to make it match better with everything already in the repo. It would probably be good to look through the commits I made to see the stylistic changes so you can keep those in mind. I also think there might have been a small problem with how the mavlink url was being sent throughout the code because the logging message displaying the url you were trying to connect to was broken. But other than that, thanks for setting this all up!

Tyler-Lentz (approved):

PR #154, opened 2024-04-17
Feat/airdrop pathing config

AskewParity:
# Pathing Configuration

- Adds some config options to Airdrop Coverage Path Planning (CPP)
  - May sweep vertically or horizontally
  - May alternate directions each passthrough or stay the same
- May choose optimal path

## Notes

- Not thoroughly tested
- Could be somewhat hardcoded to improve efficiency
- STILL
  - does NOT check for bounds
  - can NOT do non-convex shapes
  - all other listed in the original PR

### Small Changes for Future

- bounds testing
- diagonal sweeps (likely NOT useful for the competition, but this is just more math)

# OBC Config

## Changes

1. Changing nested config options to anonymous structs
    - primarily to accommodate `variables_in_snake_case`
    - looks messy, but the user will most likely never have to deal with it (especially with intellisense [which I don't have])
3. Deleted `default-config.json` 
    - as we are hardcoding the values in code
    - the extra write to the `.json` file would be unnecessary and only add possible points of failure
4. Added in `RRTConfig` as well as `AirdropSearchConfig` (might need to rename) into the config
    - were added in as structs as they are in `datatypes.hpp`


Tyler-Lentz (changes requested):
This all looks great -- thanks for doing this -- the only thing is I'm about to merge in the config json PR, so if you could go through and move these config options to the overarching config file for the repo that would be good so we don't have several configs lying around in the repo.

Tyler-Lentz (approved):
This all looks very good, thank you for going through all of the minutiae with the config and getting the airdrop pathing configurable to our needs.

The only stylistic different I would have made would have been to include the rrtconfig and coverage config in the same hierarchy as the rest of the config options, so that one word = one object in the JSON, but the way you did it probably made you modify things you already had significantly less and might be better. Either way we'll merge this in as it is now because it already does everything we need to do.

:+1:

AskewParity:
> The only stylistic different I would have made would have been to include the rrtconfig and coverage config in the same hierarchy as the rest of the config options, so that one word = one object in the JSON, but the way you did it probably made you modify things you already had significantly less and might be better. Either way we'll merge this in as it is now because it already does everything we need to do.

I can definitely see how this might make the file less readable and non-standardized as they don't follow the structure of the existing config.

However, I personally disagree, because the `pathing config` options are always used as structs, so it would be redundant to recreate the structs from a massive list of config options that just float around.

PR #155, opened 2024-04-17
Airdrop Connection Test

Tyler-Lentz:
I tested basic airdrop connection with Cody earlier in the week. This PR fixes two big bugs

## Protocol Inconsistency 

This PR fixes a bug where the airdrop boards and OBC could get out of sync if an initial ACK_MODE packet was dropped, so the board would miss the initial OBC ack and keep sending SET_MODE packets until the end of time.

## Memory access error on malformed packets

Also, this fixes a potential memory access error / seg fault by bounds checking the `data` of a heartbeat packet to make sure that it is within 1-5 before using that value to index into an array.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,11 +1,16 @@
 // See this page for reference of options: https://containers.dev/implementors/json_reference
 {
 	"name": "Existing Dockerfile",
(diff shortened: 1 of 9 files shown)


atar13 (approved):
LGTM. Nice to see this getting integrated with the airdrop hardware.

PR #157, opened 2024-04-18
2x Compile Speed up

Tyler-Lentz:
SOURCES ARE SAYING THAT THIS IS THE GREATEST OBCPP PR THEY HAVE EVER SEEN!

## Summary

Before (compiling obcpp from scratch with -j4)

image

After (still with -j4)

image

## Explanation

I did two things:

1. Use a CMake feature called Unity Builds which combines groups of files into one compilation unit. You'll notice that compiling the OBC takes significantly less steps as reported by ninja (from scratch, in the before image you'll see 103/103 vs 65/65). This is where the 2x speedup came from.
2. Use something called Precompiled headers, which precompiles commonly used headers. Unfortunately, this didn't seem to speed up compilation at all, although nor did it slow it down. I'm leaving it in because it still works with it, and theoretically someone could figure out what I did wrong and maybe get some speedup from it.

## Extra Notes:

Don't delete `dummy.c`, Dummy.


jerukan:
OH MY GOD ITS THE GREATEST PR EVER IM GONNA BECOME THE FLASH WITH THIS SPEED

AskewParity (approved):
Works in MacOS

atar13 (approved):
Works on Jetson

PR #159, opened 2024-04-23
LUCID Camera Client

### LINKED ISSUE #29: Create LUCID camera interface implementation
## 🚀 Feature
Implement a LUCID camera class following the common camera interface. 

## Motivation
We need a LUCID camera implementation.

## Pitch
"We need a LUCID camera implementation". 

## Alternatives
Don't use the LUCID camera.

### 1. ImageData
Implement the ImageData interface. This implementation will be used in future camera implementations. 

### 2. CameraConfiguration 
Implement LUCID-specific camera configuration. Try to put the most commonly used keys as their own data fields. 

### 3. CameraInterface for LUCID
Self-explanatory.  

## Testing 
Things we need to do for testing

### 1. CameraConfiguration 
Run unit tests. Test importing the file and convert it into the configuration. 

### 2. Test the converter (unit test)
The converter function within the interface is the critical component of the entire system. We should test this one thoroughly. 

### 3. Test CameraInterface
Run both unit and integration tests.

atar13:
Closes #29


Tyler-Lentz (changes requested):
Style wise everything looks good, thanks for putting this together.

I think the tests are failing however? Looks like a compile error with creating a lucid camera so maybe a test is out of date. So once we fix that and verify still works on the jetson I think we should be good

atar13:
> I think the tests are failing however? Looks like a compile error with creating a lucid camera so maybe a test is out of date. So once we fix that and verify still works on the jetson I think we should be good


Yeah that's what this comment was about. I'll try to figure it out:

https://github.com/tritonuas/obcpp/pull/159#issuecomment-2097590894

Tyler-Lentz:
> > I think the tests are failing however? Looks like a compile error with creating a lucid camera so maybe a test is out of date. So once we fix that and verify still works on the jetson I think we should be good
> 
> Yeah that's what this comment was about. I'll try to figure it out:
> 
> [#159 (comment)](https://github.com/tritonuas/obcpp/pull/159#issuecomment-2097590894)

oh sorry i didn't see that

Samir-Rashid (approved):
The code works. I'm a bit hesitant to merge because there are a lot of must fix problems left. To make sure this doesn't get even more out of hand, we should merge now. I do not want to block work on main that builds on this.

atar13:
Tests are passing now after some shenanigans with GitHub actions not detecting the environment variable for the ArenaSDK location set in the Docker image.

PR #160, opened 2024-04-23
trying to push to full image tag

atar13:
docker push action for x86 image has been broken for the past two weeks after this action failed: https://github.com/tritonuas/obcpp/actions/runs/8638682448. This was throwing a crptic error saying "buildx failed with: ERROR: unauthorized: access token has insufficient scopes". This didn't seem like a token issue since the docker login step succeeded and the push step has permissions to write to GitHub packages. This commit attempts to push an the image to the full tag "ghcr.io/tritonuas/obcpp:x86" instead of just "x86"


Samir-Rashid (approved):
LGTM

PR #165, opened 2024-04-25
fix intellisense

Tyler-Lentz:
Removes precompiled headers because these bastards break intellisense for vscode and dont give any performance benefit


Samir-Rashid (approved):
Precompiled headers sneak attack

PR #172, opened 2024-04-29
have actions run in x86

atar13:


Samir-Rashid (approved):
Good timely action fix

PR #175, opened 2024-05-03
add all supprted jetson cuda archs

atar13:
When @hashreds was testing running the saliency model we ran into this Pytorch/CUDA error. 
```
    """
    return torch.ops.torchvision.nms(boxes, scores, iou_threshold)
           ~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
RuntimeError: CUDA error: no kernel image is available for execution on the device
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

Looking online and hypothesizing makes me think that this is because torchvision in our Docker image isn't compiled to support all the architectures that the Jetson supports. I think for some reason there's some CUDA code that's trying to access stuff on an architecture that we're not compiling torchvision with.

## Solution?

I'm adding all the supported architectures on the Jetson listed by `nvcc --list-gpu-arch` to `TORCH_CUDA ARCH_LIST` while compiling torchvision to see if this resolves the issue.
image

I'm also Trojan horsing arena SDK for ARM and gdown since those were only on the camera branch.

### DIFF
--- a/docker/Dockerfile.jetson
+++ b/docker/Dockerfile.jetson
@@ -70,7 +70,8 @@ ARG TORCHVISION_VERSION=0.17.0
 # Can also run "nvcc --list-gpu-arch" on the Jetson to verify.
 # See https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#gpu-feature-list
 # and https://arnon.dk/matching-sm-architectures-arch-and-gencode-for-various-nvidia-cards/
-ARG CUDA_ARCH_LIST="8.6"
+# and https://stackoverflow.com/a/74962874
+ARG CUDA_ARCH_LIST="5.0 5.2 5.3 6.0 6.1 6.2 7.0 7.2 7.5 8.0 8.6 8.7 8.9 9.0"
 WORKDIR ${TORCHVISION_INSTALL_DIR} 
 RUN wget "https://github.com/pytorch/vision/archive/refs/tags/v${TORCHVISION_VERSION}.zip" \
     && unzip "v${TORCHVISION_VERSION}.zip" \
--- a/docker/Dockerfile.jetson
+++ b/docker/Dockerfile.jetson
@@ -47,7 +47,8 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
                            # imagemagick with c++ dependency
                            libmagick++-dev \
                            # needed for pytorch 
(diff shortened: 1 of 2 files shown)


docker/Dockerfile.jetson
```diff
@@ -47,7 +47,8 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
                            # imagemagick with c++ dependency
                            libmagick++-dev \
                            # needed for pytorch 
-                           libopenblas-dev
+                           libopenblas-dev \
+                           ninja-build
```
Samir-Rashid, line 51: How was this missing before

docker/Dockerfile.jetson
```diff
@@ -81,6 +83,18 @@ RUN wget "https://github.com/pytorch/vision/archive/refs/tags/v${TORCHVISION_VER
     && make -j4 \
     && make install
 
+RUN pip3 install gdown
+ENV PATH="${PATH}:${HOME}/.local/bin"
+
+ARG ARENA_INSTALL_DIR=/arena-tmp
+ARG ARENA_TAR_PATH="${ARENA_INSTALL_DIR}/ArenaSDK_Linux.tar.gz"
```
Samir-Rashid, line 90: Nice forward slash

Samir-Rashid (approved):
Overall good changes. Can you add && on line 60. Tyler's PR only fixed that in one of the Dockerfiles.

docker/Dockerfile.jetson
atar13: :facepalm:

Samir-Rashid:
And please fix in line 81 https://github.com/tritonuas/obcpp/blob/main/docker/Dockerfile.arm

docker/Dockerfile.jetson
atar13: :shrug: maybe we just apt installed it everytime lmao

atar13:
> Overall good changes. Can you add && on line 60. Tyler's PR only fixed that in one of the Dockerfiles.

I can't believe I missed this lmao. I was joking about having to check for &&

Tyler-Lentz (commented):
Runs on my Commodore 64!

PR #179, opened 2024-05-05
Feat/takeoff

Tyler-Lentz:
Adds functionality to take off automatically and fly the initial waypoint path

Current missing  features to be added later

- option on gcs to select manual/auto drop
- specify takeoff alt in config file
- fix altitudes of waypoints
- figure out why sitl is weird about vtol transition command


AskewParity (commented):
compiles and passes tests on my m1 macbook

AskewParity (approved):
-

Issue #176, opened 2024-05-04
BUG: THE PLANE WILL NOSEDIVE INTO THE GROUND 100% GUARANTEED!


Tyler-Lentz:
A video demonstration:

[Screencast from 05-03-2024 06:03:41 PM.webm](https://github.com/tritonuas/obcpp/assets/22576354/2ba0409d-56e6-4a4e-a177-92c56125ec05)

The root cause:

![Screenshot from 2024-05-03 18-01-03](https://github.com/tritonuas/obcpp/assets/22576354/2e4bc404-66c4-4c3a-a81e-76a2ed896437)

It just looks like we're not setting altitudes correctly in static pathing, so that should be an easy fix.

(to be clear, I think this is very funny, and a very good example of why SITL testing is important! How did all of us not notice this!)

PR #181, opened 2024-05-05
Add default takeoff altitude in config

AskewParity:
# Takeoff Altitude in Config

makes it a config instead of a magic number in `mavlink.cpp`

### DIFF
--- a/configs/config.json
+++ b/configs/config.json
@@ -10,6 +10,9 @@
             "port": 5010
         }
     },
+    "takeoff": {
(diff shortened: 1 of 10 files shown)


Samir-Rashid:
Denied. You're not matching the energy of https://github.com/tritonuas/obcpp/issues/176

Samir-Rashid (approved):
Not too familiar with this. LGTM. I'll test thoroughly during system test.

PR #183, opened 2024-05-07
Cv classification integration

shijiew555:
Changes
Made file `classification.cpp` and completed skeleton class `Classification`. Color classification has not been implemented. Made file `cv_classification.cpp` for the integration test. 

Testing
To make the target for the integration test, use `make cv_classification`. Target images for integration test is in `tests/integration/images/` and image paths could be specified in the variable `targetPaths` in `cv_classification.cpp`. The paths for model weights files for segmentation, shape classification, and character classification can be specified in variables `segmentationModelPath`, ` classificationShapeModelPath`, and `classificationCharModelPath` respectively.


CMakeLists.txt
```diff
@@ -153,6 +153,17 @@ target_add_torchvision(cv_segmentation)
...
+# cv_classification
+add_executable(cv_classification ${SOURCES} "tests/integration/cv_classification.cpp")
+target_add_json(cv_classification)
+target_add_matplot(cv_classification)
+target_add_opencv(cv_classification)
+target_add_torch(cv_classification)
+target_add_torchvision(cv_classification)
+target_add_loguru(cv_classification)
+target_add_httplib(cv_classification)
+target_add_mavsdk(cv_classification)
```
atar13, line 166: Can this be moved to `tests/integration/CMakeLists.txt`? I think your branch is out of sync with main since we recently moved all the integration tests there. But it's all good. You'll just have to merge the updated main into this branch and move this CMake snippet there.

include/cv/classification.hpp
```diff
@@ -32,6 +39,11 @@ class Classification {
     ClassificationResults classify(cv::Mat croppedImage, cv::Mat shapeMask, cv::Mat characterMask);
 
  private:
+    torch::jit::script::Module shapeModel;
+    torch::jit::script::Module characterModel;
+    torch::jit::script::Module colorModel;
```
atar13, line 44: I think you can remove the color Model private member here for now. Technically it's not a Pytorch model so we don't need to declare it with that type.

src/cv/classification.cpp
```diff
@@ -0,0 +1,109 @@
+#include "cv/classification.hpp"
+
+Classification::Classification(const std::string &shapeModelPath, \
+const std::string &characterModelPath, const std::string &colorModelPath) {
+    auto device = torch::kCPU;
+    if (torch::cuda::is_available()) {
+        device = torch::kCUDA;
+    }
```
atar13, line 8: Thank you for adding the CUDA loading :1st_place_medal:

tests/integration/cv_classification.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+        ClassificationResults results = classification.classify(target, shapeMask, characterMask);
+        std::string character = results.character;
+        std::string shape = results.shape;
+
+        std::cout << "For image " << targetPaths[i] << ": character=" << character \
+        << ", shape=" << shape << std::endl;
+
+        
+    }
```
atar13, line 49: Really like this integration test. It's nice how it runs Segmentation first since that makes it easy for us to pass in generated image from not-stolen or real images to test.

atar13 (approved):
These changes look great and the code is well organized and well documented. Thanks for taking this project on and getting the last stage of the pipeline done!

We can worry about the color model in another PR if you just want to get this merged in now. Since the color model isn't a Pytorch model, it works pretty differently than all the other models so we can talk more about it later if you want.

The one blocker before merging this in is to merge in the updated main branch into this branch. If you want, I can take care of it since I'm more familiar with the rest of the codebase. Otherwise, feel free to lmk if you need any help with the merge.

Samir-Rashid (approved):
You can merge this if you are ready. Anthony covered everything with his feedback.

PR #184, opened 2024-05-08
Chore/waypoint altitude

### LINKED ISSUE #176: BUG: THE PLANE WILL NOSEDIVE INTO THE GROUND 100% GUARANTEED!

AskewParity:
closes #176

# Path Fetching

This code's main purpose was to add altitude to the path - which it has done, but by offloading path fetching from the `RRTTree` to `RRT` static pathing.

## Justification

The path planning itself is altitude agnostic, so I've kept it that way and instead added altitude to be an extra function when a waypoint has been achieved.

This also changed the path planning to have "almost" all functionality happen between waypoints, and does not depend on the state of the tree beyond the current waypoint (the past).

## Concerns

Need to check if the flight path includes the waypoint itself, and not just the points between them.


AskewParity:
DONT MERGE, path doesn't include waypoints

Tyler-Lentz:
> DONT MERGE, path doesn't include waypoints

image

AskewParity:
Waypoints are back, but I'm still not sure if this is the correct list of points.

Tyler-Lentz:
> Waypoints are back, but I'm still not sure if this is the correct list of points.
If you want to get the OBC <--> GCS <--> SITL integration running on your localhost it might make some of the testing easier, and this is something we're going to need to do eventaully. If you're able to come down to the lab we could get that set up. I'll be here for the next 3.5 hours since Im just working on various things until the town hall

AskewParity:
The new path doesn't include the starting point, but thats probably a good thing anyways.

I'll merge the PR into main.

PR #186, opened 2024-05-10
Chore/coverage altitude

### LINKED ISSUE #185: 0 altitude in coverage path planning
## 🐛 Bug

Similar to all the other pathing options, coverage path planning is altitude agnostic, and by default zeroes out all the attitudes for the vectors generated.

## Expected behavior

Altitude should be fixed to 100 ft (30.48 m) when in 

## Possible Complications

Rate of descent may be **too steep** if we force the plane to fly to 100ft alt fro its start - provided the plane starts close to the _airdrop search zone_.

AskewParity:
closes #185 

^title

## Changes

Refactors airdrop pathing to isolate path generation from waypoint generation. 
- should be refactored again to isolate more functions, but I kinda want to move on

Adds a configurable search altitude (set by default to 30m)

Also solves the problem --> altitude is now set

## Issue

This needs to be revisited in the future, as the algorithm doesn't take in path length into account, enabling a very fast descent that may not be desirable.


Tyler-Lentz (approved):
ok but it would have been funny if the plane crashed into the ground

PR #189, opened 2024-05-24
Chore/fix config for jetson

Tyler-Lentz:


atar13 (approved):
works great. nothing sus here

PR #197, opened 2024-06-05
Prevent WaitForTakeOff from double click.

MatthewHsu1:
I was trying out GCS and OBC and realize that the takeoff button will segfault if you click it before entering the right tick or after leaving the right tick.


Tyler-Lentz (approved):
Thank you very much for making this PR. This is something I noticed a couple of weeks ago then completely forgot it was a problem. I did slightly modify your solution, however, because the root cause was that the LockPtr was incorrectly handling the error case when the tick you specify was incorrect. The optional access was supposed to catch the case when the tick was incorrect, but it wasn't. Your fix worked well, but I just made the underlying fix to the lockptr class so that the extra check isn't necessary so I removed it.

PR #198, opened 2024-06-10
Hover Search

Tyler-Lentz:
- [x] refactor config file setup
- [x] add config option to toggle hover search vs normal search
- [x] add functionality to do hover search

other things I Did

1. Package generated paths as MissionPath, which allows a generated path to be specified as a forward flight path or a hover path where each waypoint is a place at which you should hover. This class also handles converting all of the gpscoords into the appropriate mavlink mission commands
2. specify mavlink parameters in the config file which need to be uploaded to the plane
    - the current parameter is important because it makes sure the loiter mission items are handled in VTOL mode
3. added an init function for Ticks, which can handle code that should be run once when the tick is first transitioned to


tests/integration/path_planning.cpp
```diff
@@ -283,6 +283,11 @@ int main() {
 
     // RRT settings (manually put in)
     double search_radius = 9999;
+<<<<<<< HEAD
+    double rewire_radius = 256;
+    RRTConfig config = RRTConfig { num_iterations, rewire_radius, true, PointFetchMethod::Enum::NEAREST, true };
+=======
+>>>>>>> main
```
atar13, line 290: sussy

atar13 (changes requested):
lgtm. except for merge conflict

Samir-Rashid (approved):

atar13 (approved):

PR #200, opened 2024-06-11
fix: localization: resolve unit error, add real test case

Samir-Rashid:
I tested `localize` thoroughly with real test flight images
and am confident it is working correctly.

`localize` had a unit error which was causing huge amounts of
error, and there was an issue with the way tests were being asserted.
I think there are some issues with the Blender tests. There was an issue
with the `assert` that didn't catch a lat/lng mismatch since the Blender
tests are centered at (0, 0) in lat/lng. I have removed all the Blender
tests since they are not providing any value. These should be added back
for more testing at a later date.


miyatakazuya (approved):
The Blender Cases definetly needs work with creating generating better data so I'll look into our data generation another day. 

Everything with the edits seems good and the renaming adds clarity to the units that the offsets are utilizing. Glad we have real data to test on!

PR #202, opened 2024-06-12
Post request to obc - Airfrop Target 

MatthewHsu1:
Finish GSC route handler that reads in the vector of <AirdropTargets> objects and sends it to the airdrop client.


src/network/gcs_routes.cpp
```diff
@@ -260,7 +303,7 @@ DEF_GCS_HANDLE(Post, takeoff, manual) {
     LOG_REQUEST("POST", "takeoff/manual");
 
     auto lock_ptr = state->getTickLockPtr<WaitForTakeoffTick>();
-    if (!lock_ptr.has_value()) {
+    if (!lock_ptr.has_value() || state->getTickID() != TickID::WaitForTakeoff) {
```
Tyler-Lentz, line 306: this can be removed and then this pr should be ready to merge (I fixed the tick lock ptr so that it this isn't needed, since the has value call should catch if it is the wrong tick)

src/network/gcs_routes.cpp
```diff
@@ -272,7 +315,7 @@ DEF_GCS_HANDLE(Post, takeoff, autonomous) {
     LOG_REQUEST("POST", "takeoff/autonomous");
 
     auto lock_ptr = state->getTickLockPtr<WaitForTakeoffTick>();
-    if (!lock_ptr.has_value()) {
+    if (!lock_ptr.has_value() || state->getTickID() != TickID::WaitForTakeoff) {
```
Tyler-Lentz, line 318: this too

Tyler-Lentz (commented):
This is super sweet, thank you, just have a tiny thing from the merges than can be removed

ill probably handle this when its ready to merge, or if you see this before its been merged you can do it if you would like

src/network/gcs_routes.cpp
MatthewHsu1: Done

src/network/gcs_routes.cpp
MatthewHsu1: Done

Tyler-Lentz (approved):

Tyler-Lentz:
holy shit you even went back and fixed the lint :goat:

Samir-Rashid (approved):

PR #204, opened 2024-06-22
fix pipeline crashes

atar13:
- Handful of changes to handle edge cases when the pipeline is run during the fly_search tick.
- Some changes to the mock camera to make testing with predefined telemetry easier

## Testing
I tested this by running the GCS with SITL (virtual plane), OBCpp, and not-stolen server. I was able to upload a mission, have not-stolen generate the matching reference images, and have the CVAggregator run the pipeline with no crashes (albeit very slowly without CUDA support for the models). I got to end of the CVLoiter tick where all the aggregator workers completed their work.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,33 +153,39 @@ add_custom_target(pull_models
 # Saliency model
 add_custom_target(pull_saliency
   COMMAND gdown 1S1IfXlGs_pCH49DwZmbD-tZA5YH0A1gx -O ${CMAKE_BINARY_DIR}/../models/torchscript_19.pth
+  USES_TERMINAL
 )
 
 # Matching model 
 add_custom_target(pull_matching
   COMMAND gdown 1NeFiAfSSLXAZWlehfd0ox7p_jFF4YdrO -O ${CMAKE_BINARY_DIR}/../models/target_siamese_1.pt
+  USES_TERMINAL
 )
 
 
 # Segmentation model
 add_custom_target(pull_segmentation
   COMMAND gdown 1U2EbfJFzcjVnjTuD6ud-bIf8YOiEassf -O ${CMAKE_BINARY_DIR}/../models/fcn-model_20-epochs_06-01-2023T21-16-02.pth
+  USES_TERMINAL
 )
 # =============================
 
 # =============================
 # Pull testing images
+add_custom_target(pull_test_images
+  DEPENDS pull_matching_test_images pull_saliency_test_images 
+)
 
 # pull cropped images from fraternal_targets testing folder
 add_custom_target(pull_matching_test_images
-  COMMAND gdown 1opxdXw75jSQZu9s61njE6hjQkfHiKvgp -O ${CMAKE_BINARY_DIR}/../tests/integration/images/test.zip && 
-    mkdir -p ${CMAKE_BINARY_DIR}/../tests/integration/images/matching_cropped &&
-    unzip ${CMAKE_BINARY_DIR}/../tests/integration/images/test.zip -d ${CMAKE_BINARY_DIR}/../tests/integration/images/matching_cropped
+  COMMAND gdown 1vmP3HUS1SyqhdtJrP4QuFpGbyoyfaYSe --folder -O ${CMAKE_BINARY_DIR}/../tests/integration/images/matching
+  USES_TERMINAL
 )
 
 # pull cropped images from saliency images testing folder
 add_custom_target(pull_saliency_test_images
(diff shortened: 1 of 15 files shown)


Samir-Rashid (commented):
Thanks for going through this. The code looks much more robust. The fly_search state still looks a little bit risky to me. I am worried something could go wrong with the way the model is running while doing other things. Also we have not explored what could go wrong with the camera during this state.

tests/integration/cv_pipeline.cpp
```diff
@@ -35,24 +36,25 @@ const double roll = 3;
 // with an arbitrary image as input
 int main() {
     cv::Mat image = cv::imread(imagePath);
+    if (!image.data) {
+        std::cout << "failed to open testing image from " << imagePath << std::endl;
```
Samir-Rashid, line 40: Make this self documenting. Add the cmake command to get the image.

CMakeLists.txt
```diff
@@ -179,9 +179,7 @@ add_custom_target(pull_matching_test_images
 
 # pull cropped images from saliency images testing folder
 add_custom_target(pull_saliency_test_images
-  COMMAND gdown 1HJLdrm0X3VxnlQ3Z58EE9-Y5VX1NlhAz -O ${CMAKE_BINARY_DIR}/../tests/integration/images/saliency.zip && 
-    mkdir -p ${CMAKE_BINARY_DIR}/../tests/integration/images/saliency &&
-    unzip ${CMAKE_BINARY_DIR}/../tests/integration/images/saliency.zip -d ${CMAKE_BINARY_DIR}/../tests/integration/images/saliency
+  COMMAND gdown 1JvtQUroZJHo51E37_IA2D1mfdJj2smyR --folder -O ${CMAKE_BINARY_DIR}/../tests/integration/images/saliency
```
Samir-Rashid, line 182: Can you add the line `USES_TERMINAL` to each of these pull_* targets. With this you can see the standard out progress of gdown in real time.

Samir-Rashid (changes requested):

tests/integration/cv_pipeline.cpp
```diff
@@ -9,7 +9,7 @@
 // Or, any cropped not-stolen images will work
 
 // this image should be located at a relative path to the CMake build dir
-const std::string imagePath = "mock_image.jpg";
+const std::string imagePath = "../tests/integration/images/black_mountain/1718243323.jpg";
```
Samir-Rashid, line 12:
```suggestion
const std::string imagePath = "../tests/integration/images/saliency/1718243323.jpg";
```

tests/integration/cv_pipeline.cpp
```diff
@@ -9,7 +9,7 @@
...
 
 const std::string refImagePath0 = "../bin/test/test/000000910.jpg";
```
Samir-Rashid, line 14: These all also need to be fixed.
```suggestion
const std::string refImagePath0 = "../tests/integration/images/matching_cropped/test/000000910.jpg";
```

Samir-Rashid:
The pipeline looks good. Only issue I see is that I am getting `match distance -inf` for the bottle matching. I'm guessing that's just something with the test and not a real issue. Please fix the lint. It's easy to fix stuff.

CMakeLists.txt
atar13: Added in https://github.com/tritonuas/obcpp/pull/204/commits/153bdc6ef0a44cd53e7c3697c1d70f708c859706

tests/integration/cv_pipeline.cpp
atar13: Added https://github.com/tritonuas/obcpp/blob/4287233b77214752e9b35ad875f1bf74d560a29e/tests/integration/cv_pipeline.cpp#L12-L14

tests/integration/cv_pipeline.cpp
atar13: Changed in latest version.

tests/integration/cv_pipeline.cpp
atar13: Changed to an array of images at the correct path that gdown downloads them to with `ninja pull_matching_test_images`

https://github.com/tritonuas/obcpp/blob/4287233b77214752e9b35ad875f1bf74d560a29e/tests/integration/cv_pipeline.cpp#L19-L47

atar13:
> The pipeline looks good. Only issue I see is that I am getting `match distance -inf` for the bottle matching. I'm guessing that's just something with the test and not a real issue. Please fix the lint. It's easy to fix stuff.

Fixed the lint up. 

I think the -inf distance was because you ran the test with no matching references images since the paths weren't updated. Earlier, I added a check for matching to return Bottle A with infinite distance if it's given no reference images (since it has no images to compare against).

Samir-Rashid (approved):
LGTM

PR #211, opened 2024-10-14
add back integration targets

### LINKED ISSUE #209: Missing Integration Target
## 🐛 Bug

Main cmake file is missing the folder for integration tests, should be a one line fix

AskewParity:
closes #209 
why was the subdirectory removed in the first place?


miyatakazuya (approved):
Was test/integration removed earlier? I'm not sure why the prev leads did that tbh but looks good

PR #212, opened 2024-10-17
Fix unit tests using GCS Handler Macro

Tyler-Lentz:
Sets up some shared infrastructure for all of the unit tests that want to mock GCS handler requests so they can use the same macro instead of redefining it across a ton of different files.

Fixes the macro which has been breaking ever since I updated how the OBC configuration setup works.

Right now this only fixes it for the gcs_server_test unit test. Other tests should be able to run if they remove their own declaration of the macro and include the one in the common header file.

Someone else should try and apply this fix to all of the other tests that use the macro. Feel free to add onto this PR or make another one.

I can explain what is going on easier in a voice call if interested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,31 +118,11 @@ add_subdirectory(src)
 # =============================
 
 # =============================
-# Integration tests
+# Tests
 
-add_subdirectory(tests/integration)
-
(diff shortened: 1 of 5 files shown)


AskewParity (approved):
good fix

PR #218, opened 2024-12-15
updated GSD localization

sophondroplet:
Here is a summary of the changes:

1) Adjusted camera parameters to fit the raspberry pi camera

2) Added a coordinate transformation to GSDlocalization algorithm  so it rotates camera coordinates according to the heading angle of the drone. This was the issue that's causing 90 feet errors previously. 

3) Added 5 New test cases with real images taken from the first big quad test flight:

- Two test cases based on two different targets on the same image (45m altitude)
- Another two test cases on different targets on same image at around 100m altitude (100m altitude)
- One dummy test case with no offset (directly underneath)
- All images are in different quadrants in the camera coordinate space to make sure the 

Test results:
- Algorithm preforms worse a high altitudes
- Error around 10-14 feet

4) Added debugging print functions for printing GSD, camera offset values (not sure if this is the best way to debug since I'm still new to C++ and googletest)

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -11,7 +11,7 @@
             ],
             "cppStandard": "c++20",
             "intelliSenseMode": "linux-gcc-x64",
-            "compilerPath": "/usr/bin/g++"
+            "compilerPath": null
         }
     ],
     "version": 4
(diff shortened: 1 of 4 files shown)


miyatakazuya:
@sophondroplet 
I removed the deprecated tests. 
Once you address the centered image test + fix linter I'll reapprove the PR 👍

miyatakazuya (approved):
lgtm 👍

PR #224, opened 2025-01-28
Mapping

kimichenn:
- Implemented the Mapping System
- Two pass system, where the first pass "maps" the images into chunks (and can be called multiple times for new images), and the second pass "maps" the chunks into one final image.
- Integration test simulates the above scenario with two first pass calls.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 // See this page for reference of options: https://containers.dev/implementors/json_reference
 {
-	"name": "x86",
-	"image": "ghcr.io/tritonuas/obcpp:x86",
+	"name": "arm",
+	"image": "ghcr.io/tritonuas/obcpp:arm",
 	// enable when need to connect over USB to pixhawk
 	// also: need to run obcpp with sudo or add tuas user to dialout group with
 	// 	   `sudo usermod -aG dialout tuas && newgrp && bash`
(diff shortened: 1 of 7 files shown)


.devcontainer/devcontainer.json
AskewParity, line 1: ok this is going to be pretty annoying because you'll have to be more careful with your commits, but

Could you revert this file? I think it's net best to keep this default to `x86`.

kimichenn: Fixed

src/cv/mapping.cpp
```diff
@@ -0,0 +1,262 @@
...
+    cv::Mat stitched;
+    cv::Stitcher::Status status = stitcher->stitch(resized, stitched);
+    return {status, stitched};
+}
+
+std::string Mapping::currentDateTimeStr() {
+    auto now = std::chrono::system_clock::now();
+    std::time_t tt = std::chrono::system_clock::to_time_t(now);
+    std::tm local_tm;
+#ifdef _WIN32
```
AskewParity, line 83: I assume this is based on OS, we are always assuming ubuntu 22.04 I'm pretty sure, so I don't think that these preprocessor things are necessary.

AskewParity (approved):
I think this looks fine for the moment. Important things to note

1. Use Loguru instead of cout
2. Describe how to set up and run the integration test

PR #229, opened 2025-01-29
Chore/erase lucid

### LINKED ISSUE #173: Lucid Camera Client uses too much memory
## 🐛 Bug

Currently, we save all our images to a queue. Each additional image seems to be taking 100MB of memory. LUCID says the image buffer size is 20MB but in our observations memory usage is increasing much faster.

Even if the images are only taking 20MB, it will only take a few hundred images to take up all the Jetson's memory and crash it. Assuming you have 4GB free out of 8GB (since other stuff is running with obcpp or the host), it will take 200 images of size 20MB to fill that memory. This is not outside the realm of possibility to occur during flights. 

## To Reproduce

1. SSH in to Jetson
2. Checkout `feat/lucid-camera` branch
3. Set image taking delay to 0s in `camera_lucid.cpp` test to quickly take a bunch of images 
4. Compile `camera_lucid` test and run
5. Watch memory consumption skyrocket

## Potential Solutions

These aren't mutually exclusive so we can do a combination of these. 

### Option 1: Save images to disk instead of memory

Instead of saving the images to a queue, store them to a file (png or jpg; both should be less size than our memory representation). 

Pros:
- We have a lot more disk space than memory
- We can keep taking images without worrying about running out of memory
Cons:
- Disk space is slow and copying data to disk wastes CPU cycles
- Have to load image back into memory when running pipeline
- We might still run out of disk space. Right now the SD card is only 64 GB and a lot of that space is already taken up by docker images so we're running low on space. So it's not unreasonable to run out space after leaving the camera running for a while. We could mitigate this by putting back the 512GB SD card.

### Option 2: Run the pipeline sooner and free image data

Instead of waiting until all the image taking is done we can take a few images, run the pipeline sooner than later so we can free the image buffers when the pipeline (or at least the saliency stage completes). 

The implementation of this depends on how long the pipeline takes to run (which is something we have to benchmark).

- If the pipeline is quick, we can take pictures and synchronously run the pipeline, free the image and move onto taking the next one. This only works if the pipeline takes a second or two to run.
- If the pipeline takes a while, we don't want to block our image taking since we could be missing images of targets while flying over the search area. So we should have two threads running simultaneously: one taking images and another running the pipeline. This works as a producer consumer model where the pipeline consumes the images that the camera produces. The one thing to be careful of here is to make sure image producer doesn't use too much memory since it'll be producing images faster than the pipeline can finish. Also, one thing to keep in mind is the additional memory used by the pipeline (pytorch models and other stuff)

### Option 3: Optimize the existing code

There might be a way to optimize how we're storing data in the code. Maybe we can use some lossless compression instead of storing huge WxHxC arrays. It's also possible that the images are being stored multiple times somewhere as well (since it seems to be taking 100MB of data instead of 20MB). To make sure there's no memory leaks we can check with Valgrind since it seems it should work on Jetson in Docker according to the Internet.

### LINKED ISSUE #213: Fix Unit Test Failure
Currently, the Unit Tests fail because the models aren't automatically pulled.

Potential Action:
- Add feat in our cmake so that the models are pulled automatically upon build. 
- Change unit tests to not depend on models
- comment it out (oops)

### LINKED ISSUE #225: Remove Lucid Camera Code
## 🐛 Removed Hardware

- [x] #227
- [ ] #228
- [x] remove lucid code (may conflict with below)

## Other Misc Tasks associated

- [x] reconfigure github actions

## Expected behavior

The code should be identical in functionality as before, as cutting camera code will be the discretion of those who work on the interface.

### LINKED ISSUE #228: Remove ArenaSDK from dockerfile
- [x] remove references from all dockerfiles
- [x] build arm dockerfile
- [ ] build x86 dockerfile
- [x] build jetson dockerfile

AskewParity:
closes #225 #227 #228 and closes #173 by bypassing it

I have an arm docker image build and ready to push, still need jetson and x86 one.

Fixes unit tests. closes #213


dzodkin33:
https://github.com/tritonuas/obcpp/blob/main/src/camera/CMakeLists.txt

Remove `lucid.cpp` on line 6. Otherwise the compilation would fail.

AskewParity:
It compiles (I've compiled it with the new docker image). This is from the files sitll existing and the header guards in the code.

Do you think I should also just delete the lucid code from the repo in this PR or should I make that a separate branch?

dzodkin33:
@AskewParity 

I'd say delete lucid here. We're moving away from Arena and there is no point of having code that relies on arenaSDK.

dzodkin33 (commented):
lgtm

miyatakazuya (approved):
Great cleanup and works locally as well. Nice catch on the localization error and glad our tests pass (finally)

PR #230, opened 2025-02-04
Add ranked path deviation heuristic

### LINKED ISSUE #210: Waypoint Mission Coverage ability
## 🚀 Feature

Enable the waypoint mission to cover some area where we need to take pictures.

## Motivation

Whereas we have a pretty clear idea of what the mission wants the waypoints to be (corners),
Whereas our plane is pretty efficient/fast in forward flight,

we could conceivably augment path planning to deviate slightly from the most efficient path in order to take pictures.

## Pitch

We should be able to, along with hitting waypoints, fly into the coverage areas (altitude difference as well), start taking pictures, and communicate with other threads/processes. This should not dramatically compromise the complexity/efficiency of the mission, as we can always do dedicated coverage tasks (we probably will have to do at least one pass of coverage anyways).

## Alternatives

Hard Code the current algorithm for competition.

image

Above is the mapping specifications for 2025.

As seen in the image, the airdrop boundary is a subset of the mapping boundary. Because we are limited in the mission time (30 minutes), we want the ability to optimize our waypoints depending on the current state of our mapping and target detection.

## Additional context

The primary changes (most important to least)
1. Deviate a path to fly around a coverage area
  - Be able to talk to the camera (**not unique to this issue**)
  - Send messages to CV/mapping (**not unique to this issue**)
2. Record What areas have been sufficiently covered (**not unique to this issue**)
3. Path plan within the coverage zone for efficient coverage
4. Able to generate additional waypoints such that the choice for coverage is efficient.
5. Support input from CV Pipeline (TBD) but we will have a state var that waypoint can use to consider how many targets have been detected so far.
6. Support input from Mapping (TBD): Same thing, we want to know how complete our mapping is and determine if our waypoints need to be augmented.

Fox-King777:
Closes #210

Description
Goal: generate a new goals list with one deviation point to map more of the area increasing the path length as little as possible.
1. Randomly sample 50 points within mapping area
2. For each deviation point, insert it into each possible position between the goal order to create many potential new goals lists.
3. Estimate the area covered by the new path.
4. Rank the goals based on area / path length
5. Pick the best goals list

Can by ran independently from(and before) pathing and runs at 1% the time it takes for pathing.


src/pathing/environment.cpp
```diff
@@ -354,6 +356,72 @@ std::vector<XYZCoord> Environment::findIntersections(const Polygon& polygon,
...
+            }
+        }
+    }
+
+    return intersections;
+}
+
+bool Environment::isPointInMappingRegion(const XYZCoord& point) const {
+    return isPointInPolygon(mapping_region, point);
+}
```
AskewParity, line 400: I don't think this is necessary

tests/unit/environment_test.cpp
```diff
@@ -213,36 +213,74 @@ TEST(EnvironmentTest, VerticalRayIntersectsEdge) {
...
+    Polygon mapping_region2 = {{XYZCoord(0.0, 0.0, 0.0), XYZCoord(50.0, 100.0, 0.0),
+                                XYZCoord(100.0, 0.0, 0.0), XYZCoord(150.0, 100.0, 0.0),
+                                XYZCoord(200.0, 0.0, 0.0)}};
+    Environment test_env2({}, {}, mapping_region2, {}, {});
+
+    XYZCoord start4(0.0, 50.0, 0.0);
+    XYZCoord end4(200.0, 50.0, 0.0);
+    std::vector<XYZCoord> expect4 = {XYZCoord(25.0, 50.0, 0.0), XYZCoord(75.0, 50.0, 0.0),
+                                     XYZCoord(125.0, 50.0, 0.0), XYZCoord(175.0, 50.0, 0.0)};
 }
```
AskewParity, line 286: Here you should have some testing to detect literal edge cases.

Have the start or end point be on the edge and see how the code behaves. I think you may be miscalculating path length in the mapping region on these edge cases based on the interaction of `pointInPolygon()` and your calculations.

Fox-King777: Point in polygon does not include points on the edge as inside the polygon. The intersections does include the intersection points on the edges. Because of this, it does not miscalculate the estimated area. It does not miscalculate the path length because that's based only on start and end point which does not depend on the intersections.

src/pathing/environment.cpp
Fox-King777: If I refactor this, I'll have to make mapping region a public variable.

AskewParity (approved):
Thanks for doing this, the code seems to work very well. 

The tests run smoothly on my machine.

PR #231, opened 2025-02-06
Fixed CI Test Action Issue

dzodkin33:
Fixed CI not able to pull udp_squared and protos.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,10 +28,7 @@ jobs:
         run: echo "$BRANCH_NAME"
 
       - name: Clone Repo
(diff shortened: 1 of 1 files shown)


miyatakazuya (approved):
Github actions running as expected. LGTM

AskewParity (changes requested):
I don't understand what this fixes, I am aware that

- in previous commits, there was an issue with pulling these tests
- there is no issue now in the main branch, we have already messed with the GH actions

dzodkin33:
@AskewParity 

1. The error with Test CI indicated it wasn't able to pull udp_squared. 
2. Additionally, Pull Submodules became somewhat obsolete since all GitHub repos are now public.

AskewParity (approved):
Ok, I think on net this is good to simplify the code.

However, I still don't think this solves the issue (as the issue was solved already)

PR #234, opened 2025-02-22
Implement YOLO

kimichenn:
- Implement YOLO
- Added onnxruntime as a dependency for loading models.
- Integration test tests a single image and draws the bounding boxes

The models should be placed under `models/` and they can be found [here](https://drive.google.com/drive/u/0/folders/1I4SBYWlACCiR8nsaqXxLWhUFZLAqa7qN)


kimichenn:
#237

PR #236, opened 2025-02-23
Update CV Pipeline

kimichenn:
- Update CV Pipeline such that it only goes through YOLO and localization
- Remove saliency, classification, matching, and segmentation
- Remove extraneous dependencies
- Integration test simulates 1 run-through of the pipeline

You also have to comment out lines 210 and 351 in `build/_deps/onnxruntime-src/include/onnxruntime_cxx_api.h` to stop the compilation from throwing errors. (We probably should find a better solution so that we can actually run the gh tests)


kimichenn:
#237

PR #237, opened 2025-03-01
Convert wind info from mavlink passthrough

Fox-King777:
* Did not have a way to get wind data from mavlink as the conversion to struct is not officially supported.
* Use the offsets given by https://github.com/mavlink/c_library_v2/blob/master/common/mavlink_msg_wind_cov.h to get wind info


AskewParity (approved):
Looks good for now, the math checks out - we should test this in the future.

PR #248, opened 2025-04-09
Unit tests for distanceInMetersBetweenCords function. 

### LINKED ISSUE #233: Test Localization Distance Calc
Summary:
- Write tests for `distanceInMetersBetweenCords()` in localization. 

Motivation:
- I tested this function back in its python days, but I want to ensure that it has the same relative accuracy and works fine.

SofinaYang:
Fixes #233 

- Unit tests for distanceInMetersBetweenCords function in localization: 
- Set the threshold between the expected and estimated result to be 0.03
- Used data from https://www.omnicalculator.com/other/latitude-longitude-distance as reference
- Test cases include distance around thousands kilometers, around a hundred meters, and within 10 meters.

### DIFF
--- a/tests/unit/cv/localization.cpp
+++ b/tests/unit/cv/localization.cpp
@@ -213,6 +213,62 @@ TEST(CVLocalization, DistanceAccuracy) {
         },
 
         {
+            // coordinate 9 (40.459N, 50.459E)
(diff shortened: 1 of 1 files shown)


miyatakazuya (approved):
Seems like this works as intended. Nice job!

Issue #242, opened 2025-03-09
Task: Add Refueling Tick


miyatakazuya:
Add a `refueling.cpp`. 
- Should be pretty barebones, trigger state switch when GCS communicates refueling procedures are complete.

Issue #241, opened 2025-03-09
Task: Modify AirdropApproach


miyatakazuya:
## Feature 

Add logic to trigger variations of landing based on if it is on Airdrop 1 & 3, 2 & 4. Moreover:

- 1 & 3 will continue flying
- 2 & 4 triggers a switch to landing
  - 2 will enter a refuel loitering state
  - 4 will terminate the mission

kimichenn:
#250

Issue #240, opened 2025-03-09
Task: Modify Landing Tick + MissionState


miyatakazuya:
## Feature 
---
Right now landing ticks = end of mission, so the ticks do not have any capability to handle state switching.
- Modify Constructor to take in a succeeding tick
- A state switch from this tick should be triggered via a landing state boolean in mission state, as well as verifying that the landing was successful (GCS element to verify landing.)

Issue #235, opened 2025-02-22
Refueling Tick


Fox-King777:
## 🚀 Feature

After dropping the second payload, the plane will auto/manually land for refueling. During this time, the plane should be switched into a refueling tick which tells the plane to wait for a message from the GCS. The plane will be disarmed and reloaded with the last two payloads. After the GCS message, the plane will takeoff.

## Motivation

We only carry 2 of the 4 payloads at a time so we need to refuel.

kimichenn:
#250

PR #250, opened 2025-05-19
Implement Refueling + Tick restructuring 

miyatakazuya:
This PR introduces the Refueling tick, as well as overall changes to the FSM based on the new mission plan.
While it is a large PR, the changes are simple (when compared to the beast that is `feat/cv-everything`).
1. Refueling
- This tick queries the mav to check its armed status. If it is rearmed, then it signals that the team has re-armed a new payload onto the drone and so the tick escapes onto the next.
- The tick prior to this is `airdrop_approach`, and the decision to either terminate the mission/land or refuel is based on the amount of airdrops that have been dropped so far.
- The `active_takeoff` tick has also been modified to skip the CV procedure if we have done them already.
2. Other changes
- Minor changes to `static.cpp` and the searchBoundPath generation. 

Closes: #242 #241 #235 #240


src/ticks/active_takeoff.cpp
```diff
@@ -50,7 +51,14 @@ Tick* ActiveTakeoffTick::tick() {
     // NOTE: keep in sync with manual takeoff tick
     // transitions to flying waypoints tick, such that when the flying waypoints
     // tick is done it transitions to uploading the coverage path
-    return new FlyWaypointsTick(this->state, new MavUploadTick(
-        this->state, new FlySearchTick(this->state),
-        state->getCoveragePath(), false));
+
+    // This is jank as hell so make it config based plz
```
AskewParity, line 55: should this be worked on?

src/ticks/refueling.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include "ticks/mav_upload.hpp"
+#include "ticks/active_takeoff.hpp"
+#include "utilities/constants.hpp"
+
+RefuelingTick::RefuelingTick(std::shared_ptr<MissionState> state)
+    : Tick(state, TickID::Refueling) {}
+
+std::chrono::milliseconds RefuelingTick::getWait() const { return REFUELING_TICK_WAIT; }
+
+Tick* RefuelingTick::tick() {
```
AskewParity, line 13: seems redundant

src/ticks/active_takeoff.cpp
Fox-King777: Yes, I changed it, I just didn't remove the comment

src/ticks/refueling.cpp
Fox-King777: The commented code was the actual logic, we just returned the next tick for testing. I should revert it.

src/utilities/obc_config.cpp
```diff
@@ -85,6 +86,7 @@ OBCConfig::OBCConfig(int argc, char* argv[]) {
     SET_CONFIG_OPT(pathing, approach, drop_altitude_m);
     SET_CONFIG_OPT(pathing, approach, guided_drop_distance_m);
     SET_CONFIG_OPT(pathing, approach, unguided_drop_distance_m);
+    SET_CONFIG_OPT(pathing, approach, payload_size);
```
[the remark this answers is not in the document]
Fox-King777: Should we place this in takeoff configs then?

src/ticks/refueling.cpp
Fox-King777: This purpose of this tick is to wait for the plane to be armed before going to the next tick.

src/cv/pipeline.cpp
```diff
@@ -74,7 +74,9 @@ PipelineResults Pipeline::run(const ImageData& imageData) {
         GPSCoord targetPosition;
         if (imageData.TELEMETRY.has_value()) {
             targetPosition = this->gsdLocalizer.localize(imageData.TELEMETRY.value(), box);
-            std::cout << "pipeline: " << targetPosition.latitude() << " " << targetPosition.longitude();
+            std::cout << "pipeline: " <<
```
AskewParity, line 77: this seems to be a debugging statement

src/ticks/manual_landing.cpp
```diff
@@ -5,13 +5,13 @@
...
+ManualLandingTick::ManualLandingTick(std::shared_ptr<MissionState> state, Tick* next_tick)
+    :Tick(state, TickID::ManualLanding), next_tick(next_tick) {}
 
 std::chrono::milliseconds ManualLandingTick::getWait() const {
     return MANUAL_LANDING_TICK_WAIT;
 }
 
 Tick* ManualLandingTick::tick() {
-    return nullptr;
+    return next_tick;
```
AskewParity, line 16: This is an **issue**? Take the following:

1. We intend to refuel --> We set next tick to be the `refuel` tick.
2. In manual landing, it immediately transitions to `refuel` tick.
  - The safety pilot is still landing the aircraft.
  - The plane is _armed_.
3. The `refuel` tick will immediately exit, as the plane is still armed

Therefore, the arm check that refuel is meant to do is contingent on whether or not the safety pilot lands the plane in time for the first execution of the `refuel` tick.

miyatakazuya: This is an oversight on my part... 
Should have a event trigger that is David dis-arming the plane, which signals the 'end' of `manual_landing`.
I'll add that back in for clarity

PR #251, opened 2025-05-21
Change airdrop to use Ardupilot RELAY2_FUNCTION

kimichenn:
- Replaces the old UDP code with Ardupilot RELAY2_FUNCTION for dropping airdrops.
- Uses passthrough to access Mavlink directly to change the state of relay
- Integration test `drop_it_like_its_hot` tests the airdrop mechanism by calling `triggerAirdrop()`, which is defined in `airdrop_approach`


src/network/mavlink.cpp
```diff
@@ -204,9 +203,35 @@ MavlinkClient::MavlinkClient(OBCConfig config)
...
 }
 
+// Implement the triggerRelay method
+bool MavlinkClient::triggerRelay(int relay_number, bool state) {
+    if (!this->passthrough) {
+        LOG_F(ERROR, "MavlinkPassthrough interface not available");
+        return false;
+    }
+
+    // Command for DO_SET_RELAY (command ID 181)
```
AskewParity, line 213: Should we add the url `https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_RELAY` for documentation?

miyatakazuya:
omg why did the goat @atar13 return :skull:

miyatakazuya (approved):
I've tested this a couple times with the actual magnet and seems to actually work perfectly. One weird sorta related bug is that the relay is on by default, but once we send the HTTP request once, the default state is switched to off. Otherwise, looks good to merge.

atar13:
> omg why did the goat @atar13 return :skull: 

I'm locked in. I'll keep the commits coming 🫡

PR #255, opened 2025-05-30
Stream Camera

kimichenn:
Add route handlers for the gcs camera stream buttons.

### DIFF
--- a/src/network/gcs_routes.cpp
+++ b/src/network/gcs_routes.cpp
@@ -451,6 +451,49 @@ DEF_GCS_HANDLE(Post, kill, kill, kill) {
     }
 }
 
+DEF_GCS_HANDLE(Post, camera, startstream) {
+    LOG_REQUEST("POST", "/camera/startstream");
+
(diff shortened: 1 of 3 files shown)


src/network/gcs_routes.cpp
```diff
@@ -451,6 +451,69 @@ DEF_GCS_HANDLE(Post, kill, kill, kill) {
     }
 }
 
+DEF_GCS_HANDLE(Post, camera, startstream) {
+    LOG_REQUEST("POST", "/camera/startstream");
+
+    std::shared_ptr<CameraInterface> cam = state->getCamera();
+    // const string
+    uint32_t interval;
+    unsigned long parsed_ul = std::stoul(request.body); // NOLINT
```
miyatakazuya, line 460: This type casting is probably more complicated than it needs to be but works 🤷‍♂️

miyatakazuya (approved):
While it should be noted that this is only tested with the Mock, the functionalities are working as intended. LGTM

Issue #265, opened 2025-07-13
Chore: Prune Integration Tests


AskewParity:
Many of our integration tests are not necessary anymore (e.g. old CV component integration tests). This pollutes the codebase.

PR #272, opened 2025-07-26
Chore: Prune Integration Tests

### LINKED ISSUE #263: Integration Tests
Our integration tests are either broken or don't work.

AskewParity:
closes #263 #265

# Methodology

I was cautious when pruning - keeping tests that I think could give value even if they don't work at this moment.

I don't posses convictions this way or that, so if any of y'all provide warrants for erring on the side of deletion instead of preservation, I'll do it.

# Changes

I plan to squash merge these commits

- airdrop_packets (doesn't work, but compiles)
- airdrop_sockets (doesn't have compilation target)
- camera_mock 
  - change [finally] the config to look in a more accessible image directory
  - update comment about configs
- old cv stuff (deleted)
- CV - added header comments for cli args, all use png now
  - cv_aggregator
  - cv_mapping
  - cv_pipeline
  - cv_yolo 
- deviation_ranking (integration test works, algorithm doesn't work)
- drop it like it's hot
- torchvision load test (deleted --> old sv stuff)
- **Circular Dependencies** (described below)
  - plotting
  - picam

# Things I Noticed

`matplot` needs to be included for whatever reason.

## Alarming

There are come circular dependencies that can arise in our code. Specifically, `networking` and `camera` seem to be intertwined, with `camera` needing `mavlink`.


tests/integration/cuda_check.cpp
```diff
@@ -1,5 +1,9 @@
 #include <torch/torch.h>
 
+std::string bool_to_string(bool b) {
+	return b ? "true" : "false";
+}
+
 int main() {
-	std::cout << "Have CUDA? " << torch::cuda::is_available() << std::endl;
+	std::cout << "Have CUDA? " << bool_to_string(torch::cuda::is_available()) << std::endl;
```
miyatakazuya, line 8: Honestly I thing this test is also prune-able unless you've encountered scenarios where for some reason we needed to use a test to verify this instead of using `nvidia-smi`

src/camera/interface.cpp
```diff
@@ -7,6 +7,7 @@
 #include "nlohmann/json.hpp"
 #include <loguru.hpp>
 
+#include "network/mavlink.hpp"  // imported here but not in the header due to circular dependency
```
miyatakazuya, line 10: It does make sense that the camera needs mavlink but I didn't notice the circularity. This is much better for long term brevity :t-rex:

miyatakazuya (approved):
Thanks for spending the time to revisit all of the intergration tests. I think the the removed tests are all good to remove as all of them are just deprecated. I understand the PR is primarily a cleanup so I didn't spend too much time running each test, but I think overall the changes are clearer and changes to the dependencies are also a nice catch. I've left some comments about smaller details but overall I didn't see any large blockers. :100:

PR #302, opened 2025-10-12
Feat: Airdrop updates

### LINKED ISSUE #295: Update targets
Should only contain tents and mannequins as target types now

kimichenn:
Close #295 

This PR renamed all instances of `AirdropIndex` to `AirdropTarget`, as well as reduced num airdrops from 4 to 2.

Need to be on feat/cv-stuff branch protos, and run `make dev_protos`. Test with feat/rename-airdrop branch on GCS.
Note that you might want to up the num_targets from 2 to 4 for testing purposes so YOLO can detect some objects.

Test failing is expected because of protos (test passes locally).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,8 +198,6 @@ endif()
 add_custom_target(dev_protos
   COMMAND cd ${CMAKE_BINARY_DIR} &&
           cmake .. &&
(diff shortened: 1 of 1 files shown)


AskewParity (changes requested):
proto should be done first* there are some changes that would trickle down here.

AskewParity (approved):

PR #304, opened 2025-10-22
Feat/Data Aggregation

### LINKED ISSUE #303: CV Aggregation Data Structure
## 🚀 Feature

We want to store all the outputs from our cv piplines into a comprehensive data structure.

When CV aggregator returns the results, it will return a list of detected objects. We will save them into a permanent list (along with global IDs <-- protos change).

## Motivation

We want to do statistics on the CV results.

## Details:

- Modify `mission_state` to contain a map<id> of `IdentifiedTarget` protos objects.
    - Add `pruneRuns(list of runs given its ID)`: Takes in a list of runs of prune, and modifies/removes them from the list.
    - On protos, there needs to be a wrapper object that includes `IdentifiedTarget` and a 64b encoding of an image. 
- Modify `gcs_routes.cpp` in the `/target/all` route such that the map is updated when the Runs are queried. Since this may run more than once, it might need to be concatenated onto any existing route.

WilyHyperion:
closes #303, stores the results of the cv pipeline in the mission state minus the images, and implements a function to clear all runs with certain IDs

### DIFF
--- a/include/cv/pipeline.hpp
+++ b/include/cv/pipeline.hpp
@@ -3,6 +3,7 @@
 
 #include <cmath>
 #include <memory>
+#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
@@ -29,14 +30,21 @@ struct PipelineResults {
 };
 
 struct PipelineParams {
-    // Added outputPath parameter with default empty string
-    explicit PipelineParams(std::string yoloModelPath, std::string outputPath = "",
+    // yoloModelPath is optional; when absent, no CV models will be loaded.
+    explicit PipelineParams(std::optional<std::string> yoloModelPath = std::nullopt,
+                            std::string outputPath = "",
                             bool do_preprocess = true)
         : yoloModelPath{std::move(yoloModelPath)},
           outputPath(std::move(outputPath)),
           do_preprocess(do_preprocess) {}
 
-    std::string yoloModelPath;
+    explicit PipelineParams(const std::string& yoloModelPath,
+                            std::string outputPath = "",
+                            bool do_preprocess = true)
+        : PipelineParams(std::optional<std::string>(yoloModelPath), std::move(outputPath),
+                         do_preprocess) {}
+
+    std::optional<std::string> yoloModelPath;
     bool do_preprocess;
     std::string outputPath;
 };
--- a/src/cv/pipeline.cpp
+++ b/src/cv/pipeline.cpp
@@ -7,9 +7,18 @@
 
 // Pipeline constructor: initialize YOLO detector and the preprocess flag.
 Pipeline::Pipeline(const PipelineParams& p)
-    : yoloDetector(std::make_unique<YOLO>(p.yoloModelPath, 0.05f, 640, 640)),
-      outputPath(p.outputPath),
-      do_preprocess(p.do_preprocess) {}
+    : outputPath(p.outputPath),
+      do_preprocess(p.do_preprocess) {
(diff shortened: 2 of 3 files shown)


miyatakazuya:
OK. Can you make a corresponding PR in the protos repository? I'll manually verify the tests since we shouldnt modify the master CI.

include/core/mission_state.hpp
```diff
@@ -113,6 +113,13 @@ class MissionState {
     OBCConfig config;
 
     std::optional<airdrop_t> next_airdrop_to_drop;
+    // removes all runs whose id is inside removals
+    void pruneRuns(std::vector<int> removals);
```
AskewParity: pruning doesn't necessarily entail deletion, I image that we may need to reclassify the objects.

include/core/mission_state.hpp
```diff
@@ -143,6 +150,11 @@ class MissionState {
     // with the detected_target specified by the index
     std::array<size_t, NUM_AIRDROPS> cv_matches;
 
+    std::mutex aggregated_data_mut;
+    // record of all cv outputs
+    std::shared_ptr<std::vector<CVResultRecord>> aggregated_data;
+    // If we've pruned runs yet. If true, then no new data will be added to the aggregated_data list
```
AskewParity: 1. We have a custom `LockPtr` class that abstracts this. Just for consistency in the codebase, this should be a `LockPtr`.
2. Since we don't care about order, we should make this a map based on id.

include/core/mission_state.hpp
```diff
@@ -143,6 +150,11 @@ class MissionState {
...
+    bool havePrunedData;
```
AskewParity: We will be concurrently pruning and adding data to the set. The images will come down to the GCS while the plane is still taking pictures.

AskewParity (changes requested):
Thanks for doing this. The general idea encapsulates what we want.

However, I failed to communicate some quirks in our codebase to you (I left some comments). Additionally, I misunderstood the conflict between records and gcs objects - I'll put some comments on the protos side for that.

include/core/mission_state.hpp
miyatakazuya: Agreed.

@WilyHyperion MB i shouldve told u to use lockptr when u asked

src/network/gcs_routes.cpp
```diff
@@ -412,8 +410,6 @@ DEF_GCS_HANDLE(Get, targets, all) {
 
         // END COMPRESSION
 
-        target.set_picture(b64);
-
```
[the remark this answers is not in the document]
miyatakazuya: Agreed

miyatakazuya:
I think alot of the comments that Chris addresses are implementation dependent and contradicts the issue details so you can hold off on the comments about `pruneRuns` and the concurrent adding/removing.

@AskewParity We should be on the same page about the implementation assigning the tasks (at least what's written on the issues) since I don't want people to be following issues guidelines but then being told it's incorrect.

WilyHyperion:
Looking back over this, I think it might be better for me to reset this and try again from scratch once I better understand the intention.  All of the code is going to have to be rewritten anyway, so it'll probably be slower and messier to try to salvage the current implementation.

WilyHyperion:
This is good for review - the 1 test failing happens because the model doesn't exist, but I need to create a data aggregator object to test parts of this. I think there was some mention of testing the whole "/targets/all" route, which isn't in this PR. If needed, I could add that

PR #311, opened 2025-11-01
Fix aggregator integration test compilation Add ninja and obcpp to CI test

kimichenn:
Aggregator integration test should compile now (ie, running `ninja` shouldn't throw any more errors).

It also turns out we don't run `ninja obcpp` or `ninja` at all in our CI tests, which is a bad thing. These are added now.


miyatakazuya (approved):
LGTM 👍

PR #317, opened 2025-11-16
Feat/clustering integration

### LINKED ISSUE #316: Integrate Clustering Into Pipeline
## 🚀 Feature
Apply the clustering algorithm to the results of the CV algorithm,

WilyHyperion:
closes #316 
integrate the clustering code into the pipeline, and use the clustering result as the airdrop target. Also includes some minor changes to the unit test from the last pr

### DIFF
--- a/src/ticks/cv_loiter.cpp
+++ b/src/ticks/cv_loiter.cpp
@@ -40,37 +40,7 @@ Tick* CVLoiterTick::tick() {
             AirdropType::Water, AirdropType::Beacon,
         };
         */
-
-        LockPtr<std::map<int, IdentifiedTarget>> results = state->getCV()->getCVRecord();
-        std::vector<std::vector<GPSCoord>> cluster_input;
-        for (const auto& pair : *results.data.get()) {
-            auto cords = pair.second.coordinates();
-            auto types = pair.second.target_type();
-            for (int i = 0; i < cords.size(); i++) {
-                GPSCoord location = cords.at(i);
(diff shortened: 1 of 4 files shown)


src/ticks/cv_loiter.cpp
```diff
@@ -40,7 +41,35 @@ Tick* CVLoiterTick::tick() {
...
+
+        LockPtr<MatchedResults> matched = state->getCV()->getMatchedResults();
+        std::unordered_map<AirdropType, AirdropTarget> matched_clusters;
+        for (int i = 0; i < clusterCenters.size(); i++) {
+             AirdropTarget airdrop;
+             airdrop.set_index(static_cast<AirdropType>(i));
+             airdrop.mutable_coordinate()->CopyFrom(clusterCenters[i]);
+             matched_clusters.insert(std::pair(static_cast<AirdropType>(i), std::move(airdrop)));
+        }
+        matched.data->matched_airdrop = std::move(matched_clusters);
```
AskewParity: This work should probably done by the `Aggregator` object. If you notice, you are getting pointers to the `Aggregator` variables and working on them here.

It would be cleaner for that work to be done in `Aggregator` and for a single function call to live here (basically telling `Aggregator` that we are ready to move on).

AskewParity (changes requested):
Sorry for the late response. I haven't tested this yet, and will do later, I just have a comment about codebase management

PR #318, opened 2025-11-16
jsonable interface

2q46:
A jsonable class for classes to inherit from used to inherit the "to_json" method which serializes them into JSON format.


include/utilities/jsonable.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+using json = nlohmann::json;
+
+class jsonable {
+
+    json to_json();
+
+};
+
+#endif
```
AskewParity: add the comment (for consistency)

AskewParity (changes requested):
Thanks for doing this. This should serve as a starter contract for how we serialize objects. The next steps should be to convert more and more things into jsonable (if we want to look at them for debugging) - although I heard that you wish to move to `not-stolen`, which is probably for the best (this stuff really only needs to be ready for real test flights). 

Note your code doesn't compile in CI. I saw it compiled locally, so I'll check it out locally, but that is something that shouldn't happen.

PR #319, opened 2025-11-22
Update YOLO Preprocessing

### LINKED ISSUE #277: Implement Owlv2
From brief testing via Python, it seems like Owlv2 detects objects much better than YOLO, so we should look into implementing it here. I have no idea as to how the logistics for this will work, but I'm guessing it's something along the lines of downloading the weights from HuggingFace, then loading them manually with torch.

kimichenn:
closes #277 

Update some yolo stuff to support the new structure of the model.

Notable changes:
- Detection threshold bumped to 0.35 (from 0.05)
- Threshold is now a setting in config jsons instead of scattered everywhere (the only exception is unit/integration tests, as I assume we want to mess with this number often)
- Added NMS to deduplicate bboxes
- Changed default image size from 640x640 to 1024x1024 (I tried both image sizes for training/inference, and the 1024 works better so we'll stick with that)

The model can be found [here](https://drive.google.com/file/d/19QFQTaGz4cQUztIb81395fido-zOJuGa/view?usp=drive_link).

I don't foresee us using the old model, so this can be merged regardless of whether I make future updates to the model (the architecture should be the same), with a big caveat, as detailed below. I think we'll just stick to finetuned YOLO this year.

**The caveat**: it is important to note that this model does NOT work for any not-stolen images (this might change after we update not-stolen to only use human and tent images). This means that we cannot run the full sitl suite unless you manually add detection yourself in the reports page.

### DIFF
--- a/include/cv/yolo.hpp
+++ b/include/cv/yolo.hpp
@@ -29,8 +29,8 @@ class YOLO {
      * @param inputHeight Height of the model input
      * @param nmsThreshold IoU threshold for Non-Maximum Suppression
      */
-    YOLO(const std::string& modelPath, float confThreshold, int inputWidth = 1024,
-         int inputHeight = 1024, float nmsThreshold = 0.9f);
+    YOLO(const std::string& modelPath, float confThreshold, int inputWidth,
+         int inputHeight, float nmsThreshold = 0.9f);
 
     /**
      * @brief Destroy the YOLO object


include/cv/yolo.hpp
```diff
@@ -74,15 +74,16 @@ class YOLO {
      * @param color     Letterbox color fill (default 114 for typical YOLO)
      * @return cv::Mat  The letterboxed image of size [newHeight x newWidth]
      */
-    cv::Mat letterbox(const cv::Mat &src, int newWidth, int newHeight,
-                      const cv::Scalar &color = cv::Scalar(114, 114, 114));
+    cv::Mat letterbox(const cv::Mat& src, int newWidth, int newHeight,
+                      const cv::Scalar& color = cv::Scalar(114, 114, 114));
```
AskewParity: I think this level of documentation is sufficient for the magic numbers

src/cv/yolo.cpp
```diff
@@ -1,16 +1,70 @@
...
+#include <algorithm>
 #include <atomic>
+#include <unordered_map>
 
 // For simplicity, we are not doing advanced error handling
 // Make sure to catch and handle exceptions in production code
 
-YOLO::YOLO(const std::string& modelPath, float confThreshold, int inputWidth, int inputHeight)
+namespace {
+float computeIoU(const Detection& a, const Detection& b) {
```
AskewParity: Documentation

AskewParity (changes requested):
I haven't actually looked to test this model on sitl, but the detection images you sent me look very promising.

AskewParity (approved):

PR #321, opened 2025-12-05
Qol/fall 2025

AskewParity:
# Description

Does a bunch of Quality of Life things

# Changes

## Done

- [x] fixed compilation warnings
- [x] added vscode setting in devcontainer that formats to cpp-lint specification
- [x] proofread (roughly README)
- [x] change ninja default core usage

## TODO

- [/] fix `run sitl-compose`
  - this is a `not-stolen` issue
- [-] prune README (it's too long)
  - defet this to another time


kimichenn (approved):
Thanks for doing this. ninja has never looked this clean before

PR #322, opened 2026-01-17
Compression Algorithm

2q46:
Compression algorithm added in cv utilities. Fixed the compression algorithm code so it is more readable and also added some image preprocessing.

### DIFF
--- a/include/cv/utilities.hpp
+++ b/include/cv/utilities.hpp
@@ -29,14 +29,16 @@ struct DetectedTarget {
 // Helper function to crop out a bounding box from an image if you still need it
 cv::Mat crop(const cv::Mat& original, const Bbox& bbox);
 
-/*
-Denoises and compresses a JPEG image to a desired quality.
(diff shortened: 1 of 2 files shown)


include/cv/utilities.hpp
```diff
@@ -27,4 +30,6 @@ struct DetectedTarget {
 // Helper function to crop out a bounding box from an image if you still need it
 cv::Mat crop(const cv::Mat& original, const Bbox& bbox);
 
+std::optional<cv::Mat> compressImg(const cv::Mat& img, int quality = 60);
```
AskewParity: Although pedantic, I think it's good to have documentation through a header comment. For instance, it also looks like we are denoising?

AskewParity (changes requested):
Starting at `gcs_routes.cpp:358` there is compression code that doesn't use your utility.

I think besides the lint errors, this is well written. Thanks for doing this.

AskewParity (approved):

PR #326, opened 2026-01-26
Pi-camera UDP Client + boost MVP

### LINKED ISSUE #214: Raspberry Pi HQ Camera Client
**Summary**: 
The Raspberry Pi HQ Camera (or global shutter version) needs to be implemented to replace the existing Lucid Camera Client.

**Features to develop**
1. ImageData
_Implement the ImageData interface. This implementation will be used in future camera implementations that require image-associated telemetry._

2. CameraConfiguration
_Implement pi camera configuration. Useful for camera config dependent features (ex. localization)_

3. CameraInterface 
_Self-explanatory._

**Testing**
1. Implement Unit Tests
_Test if client works independently to properly verify connection and perform standard client features_
2. Implement Integration Tests
_Test if the interface works with interfacing with Mapping + CV pipelines_

**Note**:
I think who ever is working on the task should look at the existing framework so that the interfaces are kept but modified to work with Picam. They should also develop a version of the MockCamera for implementation testing which fetches images from test folders instead of needing to connect and take pictures with the Camera. 

- [x] #222
- [ ] #223

### LINKED ISSUE #291: Prune/Archive Picamera.cpp
This thing is so bad. I uses some random gstreamer command to jerry rig a stream and capture an image. I hate this. This will definetly be deprecated with the new camera client (PLEASE) so either prune/archive. Probably prune

### LINKED ISSUE #331: Add camera-client as a configrable camera-type
The `config.camera.type` allows us to configure which camera is being used at runtime in `obc.cpp`. This feature current does not support the new udp-client. On top of that, the camera intrinsics + serveral other configs are hardcoded in `rp_connection.hpp`.

TODO:
- Work off branch `/feat/pi-client`
- Prune `rp_connection.hpp` to avoid hard-coding of configs (resolution, buffer, ip, port, etc.)
    - Some stuff will be static/constant so they can be hardcoded elsewhere (headerSize, magic, chunk, header struct, etc.)
- Modify `rpi.cpp` to utilize `CameraConfig` over hard-coded values in rpi_connection.hpp

miyatakazuya:
Closes #214  Closes #291 Closes #331

## Changes

This PR introduces the MVP for the Pi-Camera Client. It includes several classes, mainly `rpi.cpp/hpp` and `udp_client.cpp/hpp`. The former implements the `CameraInterface` spec, and the latter implements the `UDPClientInterface` based off how the `camera-things` project sends images. 

Here is a quick explanation of the request lifecycle between the `obcpp` RPi client and the `camera-things` server:

1. **Request Initiation**: The client (`rpi.cpp`) sends a single ASCII character `'I'` (Request Image) within a UDP packet to the `camera-things` server port.
2. **Image Capture**: The server processes the request and prepares the image as raw YUV420 planes (Y, U, V).
3. **Plane Transmission (Header)**: For each of the three planes sequentially, the server sends a packed binary header packet over UDP. This header contains a `magic` number (`0x12345678`) for validation, the `total_chunks` that will follow, and the `mem_size` (total size of the plane in bytes).
> This has shown some undterministic behavior, so looking at server for better magic number and chunking strategy.
4. **Plane Transmission (Chunks)**: The server then transmits the plane data split into multiple chunk packets to fit within network MTU limits (typically ~1024 bytes data chunks). Each chunk is prefixed with a 4-byte sequential chunk index in network byte order.

5. **Reconstruction**: The client attempts to receive the header, validates the `magic` number, and reconstructs the plane by receiving the specified number of data chunks. This process is repeated for all three planes (Y, U, and V).
6. **Formatting**: Once all three planes are fully received, `rpi.cpp` strips any padding based on specified image strides, aggregates them into a contiguous I420 buffer, and converts it to a standard OpenCV `cv::Mat` (BGR format) for further processing.

The RPI Client has several pre-defined behaviors to handle UDP unreliability:
- Times out request if frames or chunks are not received after defined timeout duration.
- Checks request magic, total chunks, and chunk size for data integrity.

On top of these additions, I've pruned deprecated `PiCamera.cpp` from 2024 comp which utilized gstreamer. 
- Also removed any remaining lucid mentions

## Testing

The main testing solution is `camera_integration_test.cpp`. It allows you to test the client, assuming that the pi-client is active.

The client/server lacks unit testing, so future issues will address implementation of the udp_server as a mock pi-server for better testing.

## Future Improvements

- Add a more robust error handling and recovery mechanism
- Add a more robust chunking mechanism
- Add a more robust magic number strategy
- Add a more robust timeout policy. (Some edges cases)
- Add more configuration options for UDP IP, Port, image res, timeout, all other stuf
- Implement Better Unit Testing

## Feedback

Any feedback is appreciated, but focusing on networking constructs and testing stability.
- Approaches on how to better maintain this type of code that is harder to test is appreciated. 
- Any other C++ standards/practices that i may have ignored. 

> Shouthout @smhitle and @dzodkin33 for the groundwork, and cse124 for helping me survive UDP nonsense


deps/boost/boost.cmake
AskewParity, line 1: The long term solution is to add this to the docker-file correct? If that is the case, I should start looking into implementing it.

include/camera/rpi.hpp
```diff
@@ -0,0 +1,63 @@
...
+using json = nlohmann::json;
+using namespace std::chrono_literals; // NOLINT
+
+namespace asio = boost::asio;
+
+const std::uint8_t START_REQUEST = 's';
+const std::uint8_t PICTURE_REQUEST = 'I';
+const std::uint8_t END_REQUEST = 'e';
+const std::uint8_t LOCK_REQUEST = 'l';
+
```
AskewParity: These should be enums

include/core/mission_state.hpp
```diff
@@ -143,6 +145,8 @@ class MissionState {
     // with the detected_target specified by the index
     std::array<size_t, NUM_AIRDROPS> cv_matches;
 
+    // boost::asio::io_context io_testies;
+
```
AskewParity: Commented code should probably be deleted

include/network/rpi_connection.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+const int headerSize = 12;
+const uint32_t EXPECTED_MAGIC = 0x12345678;
+const size_t CHUNK_SIZE = 1024;
+
+struct Header {
+    uint32_t magic;
+    uint32_t total_chunks;
+    uint32_t mem_size;
+};
```
AskewParity, line 30: Should this be in config?

include/network/serialize.hpp
```diff
@@ -0,0 +1,28 @@
...
+#define INCLUDE_NETWORK_SERIALIZE_HPP_
+
+#include <boost/asio.hpp>
+
+/*
+    Includes some reference on how to serialize and deserialize structs
+*/
+
+namespace serialh {
+    // TODO: make this so that it returns any struct
```
AskewParity: These TODOs seem to be accomplished?

include/network/udp_client.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "rpi_connection.hpp"
+
+namespace asio = boost::asio;
+
+class UDPClient {
+ private:
+        asio::ip::udp::socket socket_;
+        std::string ip;
+        int port;
+        int current_timeout_ms_ = 2000;
```
AskewParity, line 20: Config? Or does this need to be known at compiletime/before configurations are set?

src/camera/interface.cpp
```diff
@@ -17,81 +17,77 @@ CameraInterface::CameraInterface(const CameraConfig& config) : config(config) {}
...
+        return {};
+    }
+
+    auto [lat_deg, lon_deg] = mavlinkClient->latlng_deg();
+    double altitude_agl_m = mavlinkClient->altitude_agl_m();
+    double airspeed_m_s = mavlinkClient->airspeed_m_s();
+    double heading_deg = mavlinkClient->heading_deg();
+    double yaw_deg = mavlinkClient->yaw_deg();
+    double pitch_deg = mavlinkClient->pitch_deg();
+    double roll_deg = mavlinkClient->roll_deg();
```
AskewParity, line 55: see! we do use telemetry! Also this implies that our localization code has been broken since 2024.

src/camera/interface.cpp
```diff
@@ -17,81 +17,77 @@ CameraInterface::CameraInterface(const CameraConfig& config) : config(config) {}
...
+                          .altitude_agl_m = altitude_agl_m,
+                          .airspeed_m_s = airspeed_m_s,
+                          .heading_deg = heading_deg,
+                          .yaw_deg = yaw_deg,
+                          .pitch_deg = pitch_deg,
+                          .roll_deg = roll_deg};
 }
 
 bool ImageData::saveToFile(std::string directory) const {
+    if (this->TIMESTAMP == 0) {
```
AskewParity: Is this always true if the image is empty?

src/camera/interface.cpp
```diff
@@ -17,81 +17,77 @@ CameraInterface::CameraInterface(const CameraConfig& config) : config(config) {}
...
+        LOG_F(INFO, "Saving image to %s", img_filepath.string().c_str());
+        LOG_F(INFO, "Saving telemetry to %s", json_filepath.string().c_str());
+
         saveImageToFile(this->DATA, img_filepath);
         if (this->TELEMETRY.has_value()) {
             saveImageTelemetryToFile(this->TELEMETRY.value(), json_filepath);
         }
     } catch (std::exception& e) {
         LOG_F(ERROR, "Failed to save image and telemetry to file");
+        // std::cout << "Failed to save image and telemetry to file" << '\n';
```
AskewParity, line 89: delete comment

src/camera/rpi.cpp
```diff
@@ -0,0 +1,183 @@
...
+// Setup Logging
+// ...
+
+RPICamera::RPICamera(CameraConfig config, asio::io_context* io_context_)
+    : CameraInterface(config), client(io_context_, SERVER_IP, SERVER_PORT) {
+    // this->connected = false;
+    LOG_F(INFO, "RPICamera exists");
+}
+
+void RPICamera::connect() {
```
AskewParity, line 24: Should this return a bool to inform whether or not the camera connected, or is `client.connect()` blocking until it connects?

src/camera/rpi.cpp
```diff
@@ -0,0 +1,183 @@
...
+    auto start_time = std::chrono::steady_clock::now();
+
+    // 1. Send Request
+    if (!client.send(PICTURE_REQUEST)) {
+        LOG_F(ERROR, "Failed to send picture request");
+        return {};
+    }
+
+    // 2. Read 3 Planes
+    std::vector<std::vector<uint8_t>> planes = readImage();
```
AskewParity, line 59: Is it possible to have a race condition with how headers are sent and processed?

README.md
```diff
@@ -123,6 +123,47 @@ The airdrop payloads are currently programmed to connect to a WIFI network with
 this information on boot. To make sure that it is broadcasting correctly, you
 can try and connect to it via your phone or other device.
 
+## Jetson Setup (PiCamera)
+> How to setup networking for a new Jetsoj
```
AskewParity, line 127: Jetsoj

src/network/udp_client.cpp
```diff
@@ -0,0 +1,159 @@
...
+               sizeof(read_timeout));
+
+    LOG_F(INFO, "Connected to %s on port %d", this->ip.c_str(), this->port);
+    return true;
+}
+
+void UDPClient::setReceiveTimeout(int timeout_ms) {
+    this->current_timeout_ms_ = timeout_ms;
+    struct timeval read_timeout;
+    read_timeout.tv_sec = timeout_ms / 1000;
```
AskewParity, line 36: does this truncate milliseconds? (i.e. would 999ms --> 0s)

src/network/udp_client.cpp
```diff
@@ -0,0 +1,159 @@
...
+    int bytesRead = this->socket_.receive_from(asio::buffer(&header, sizeof(Header)),
+                                               sender_endpoint, 0, ec);
+
+    if (ec) {
+        LOG_F(ERROR, "Failed to read header: %s", ec.message().c_str());
+        return {};
+    }
+
+    // Network to Host Byte Order conversion handled by caller or here?
+    // rpi.cpp expects to handle it, but todo
```
AskewParity: I assume this is already handed by the caller?

To answer where it should be handled, I think it is wherever the header is processed so that there isn't a arbitrary division of labor in processing.

src/network/udp_client.cpp
```diff
@@ -0,0 +1,159 @@
...
+    std::vector<std::uint8_t> buf(bufSize);
+    std::vector<bool> received_chunks(total_chunks, false);
+
+    int chunks_received_count = 0;
+
+    // Use raw buffer? hopefully no mem leaks
+    char chunk_buf[CHUNK_SIZE + sizeof(uint32_t)];
+
+    // Loop until all chunks received
+    // Note: we prolly want to add a max retry count or timeout check inside the loop
```
AskewParity, line 122: We should. 

> I don't think we ever experienced the obc encountering an infinite loop here; although, it may in the future if we don't resolve it.

src/network/udp_client.cpp
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        if (bytesRead < sizeof(uint32_t)) continue;
+
+        uint32_t chunk_idx = ntohl(*reinterpret_cast<uint32_t*>(chunk_buf));
+        size_t data_size = bytesRead - sizeof(uint32_t);
+        size_t offset = chunk_idx * CHUNK_SIZE;
+
+        if (offset + data_size > static_cast<size_t>(bufSize)) {
+             LOG_F(ERROR, "Chunk exceeds buffer bounds");
```
AskewParity, line 146: Is this from a mis-sent chunk, or something in processing failing?

src/network/udp_client.cpp
```diff
@@ -0,0 +1,159 @@
...
+             continue;
+        }
+
+        // copy into buffer
+        memcpy(buf.data() + offset, chunk_buf + sizeof(uint32_t), data_size);
+
+        if (!received_chunks[chunk_idx]) {
```
AskewParity, line 153: What does it mean if `received_chunks[chunk_idx]` is not `true`? Are chunks sent multiple times for redundancy, or could there be a processing error?

AskewParity (changes requested):
BIG! Thanks for doing this. I think the picam is a platform we will use for a decent while, so this not only is important for near term testing, but also will be the infrastructure for future camera stuff.

Some of the comments I left are more facetious than others. Overall, I don't have a lot of constructive comments to provide, but I have left some questions about the meaning of certain control flow statements.

deps/boost/boost.cmake
miyatakazuya: Yes I think we can look into using `libboost-dev` from the apt repository instead.

include/network/rpi_connection.hpp
miyatakazuya: I think it depends, I can add all of these to a specific PiCamera config but it would result in a picamera-specific config that isn't used for mock or anywhere else. I plan to implement a feature where certain network intrinsics are configured on the `camera-things` side, and fetched via UDP instead of being hardcoded. I think for the MVP this is fine, but if you think a config is needed now I can start working on that in this PR.

include/network/serialize.hpp
miyatakazuya: Will mark as completed

src/camera/interface.cpp
miyatakazuya: 😭 we should address in different PR but oh lord

README.md
miyatakazuya: increases our non-vibecode rep AGI MVP SOP

src/camera/interface.cpp
miyatakazuya: Yes the TIMESTAMP null if its empty, but ill change it to check DATA instead for clarity.

src/camera/rpi.cpp
miyatakazuya: Since there is no notion of a connection handshake in UDP, so unless we want to implement a heartbeat to the PI so the `connect()` would probably remain a void function otherwise

include/camera/rpi.hpp
miyatakazuya: will implement

src/camera/interface.cpp
AskewParity: It's already fixed in PR #348

src/network/udp_client.cpp
miyatakazuya: Yeah, that makes sense. Moving the conversion into `recvHeader` groups the header parsing logic so the caller doesn't have to worry about the byte order. will change

src/network/udp_client.cpp
miyatakazuya: Since UDP has no packet safety, it's possible for packets to duplicate if the server assumes they were incorrectly dropped. The check is to track the number of unique packets that are processed. If `recieved_chunks` is true, then it means we've received a duplicate and so we discard it

src/network/udp_client.cpp
miyatakazuya: This is just a standard safeguard for malformed UDP packets from some network loss. If the chunk_idx or data_size is corrupted, it could result in a out-of-bounds memory write in `memcpy` so its just a safety measure

src/network/udp_client.cpp
miyatakazuya: I can implement a MAX iteration for the loop but we have also never seen the behavior yet

src/network/udp_client.cpp
miyatakazuya: it does not

src/camera/rpi.cpp
miyatakazuya: I think it is possible that we recieve a data chunk before the metadata header if it was dropped, which might result in a race condition. I think this could be resolved by adding a packet-type identifier, but this change would need to be made on both projects... I think this leads to wanting something protos but I dont know if we want to get that far yet.

miyatakazuya:
@AskewParity I've addressed the ENUM, byte order, and pruned the misc comments.
I've addressed all the comments in some form, so let me know any additional feedback

AskewParity (approved):

PR #338, opened 2026-02-12
Feat/pics during laps

### LINKED ISSUE #329: Take pictures when in the bounds of airdrop zone
## 🚀 Feature

When the plane detects it is in the airdrop zone, start taking pictures and running 
CV pipelines.

## Motivation

It is a possibility that we may fly into/near the airdrop zone while doing laps.
It may be worthwhile to take pictures when we overlap (for more detections).

## Pitch

There should be a thread that runs periodically (maybe every 300 ms) that checks for
whether or not the GPS location is within airdrop bounds. If it is, then we start
taking pictures - else we stop taking pictures.

aaryan-1310:
Closes #329 
When you fly over the search area when doing laps, take pictures and run the CV pipeline on them

### DIFF
--- a/include/ticks/fly_waypoints.hpp
+++ b/include/ticks/fly_waypoints.hpp
@@ -24,6 +24,7 @@ class FlyWaypointsTick : public Tick {
 
  private:
     Tick* next_tick;
+    std::chrono::milliseconds last_photo_time;
 };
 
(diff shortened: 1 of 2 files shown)


kimichenn (approved):
I just realized this was never reviewed/merged. Tested on full sitl suite lasts week and everything works as expected.

PR #340, opened 2026-02-19
Feat/scale env bounds

2q46:
Added the function to scale the environment bounds by a fixed amount. In the OBC config file, I set the offset to 10 but I am unsure if this is correct because I do not know the units for XYZCoord.


miyatakazuya:
does this need review? If so please request review @2q46

PR #346, opened 2026-02-26
Fix: Pathing Inputs

### LINKED ISSUE #341: Pathing from home
# Issue

The pathing code is written to where the first waypoint is assumed to be the plane's current location (or home location).

However, this is not provided in the waypoint list by default, leading to possibly either inefficient flight by ardupilot.

# Solution

1. Put the home position in the waypoint mission
    - human enforce (bad)
    - make GCS do it (less bad)
2. Always pull the plane's current location using mavlink to be the first waypoint.

AskewParity:
**MUST MERGE MAV UPDATES BEFORE THIS**

closes #341

- airdrop pathing now starts from current plane location
- waypoint pathing starts from current plane location
    - waypoint pathing is now dynamic, with a new path generated after each lap finishes.
- adds a helper function in `static` to get current plane location

# Testing

- run in SITL (not stress tested, just generic routes)
- passes pathing integration tests (`path_planning`, `coverage_pathing`, `airdrop_approach` <-- terrible naming scheme)

# TODO

- Issue #350 : coverage pathing should also by dynamic
- ticks should probably be restructured for cleaner code
- pathing performance needs to be benchmarked and optimized (rewritten lol).
- If the jetson can not make calculations quickly enough, a bandaid fix is to delete the first few waypoints in pathing under the assumption that we've already crossed them. (risk is that on a turn we have a waypoint).


src/pathing/static.cpp
miyatakazuya: should modify/remove these comments

src/ticks/fly_waypoints.cpp
```diff
@@ -60,15 +73,39 @@ Tick* FlyWaypointsTick::tick() {
...
 
-    if (isMissionFinished) {
-        if (state->config.pathing.laps > 1) {
-            state->config.pathing.laps--;
-            state->getMav()->startMission();
+    if (!isMissionFinished) {
+        return nullptr;
+    }
+
+    state->config.pathing.laps--;
```
miyatakazuya: I think its bad practice to modify anything from the config period(should just be const tbh). I think the current_lap should be mutable only via Mission_state

src/pathing/static.cpp
```diff
@@ -663,11 +665,9 @@ MissionPath generateInitialPath(std::shared_ptr<MissionState> state) {
 MissionPath generateSearchPath(std::shared_ptr<MissionState> state) {
     std::vector<GPSCoord> gps_coords;
     if (state->config.pathing.coverage.method == AirdropCoverageMethod::Enum::FORWARD) {
-        RRTPoint start(state->mission_params.getWaypoints().front(), 0);
-
-        // TODO , change the starting point to be something closer to loiter
-        // region
+        RRTPoint start(state->mission_params.getWaypoints().back(), 0);
```
miyatakazuya, line 662: Is there a rationale to changing the coverage path to generate from the end of the waypoint mission?

src/ticks/fly_waypoints.cpp
```diff
@@ -60,15 +73,39 @@ Tick* FlyWaypointsTick::tick() {
...
+        const int wait_time_ms = 100;
+
+        while (init_status != std::future_status::ready && count_ms > 0) {
+            LOG_F(WARNING, "Waiting for path to be generated...");
+            std::this_thread::sleep_for(std::chrono::milliseconds(wait_time_ms));
+            init_status = init_path.wait_for(std::chrono::milliseconds(0));
+            count_ms -= wait_time_ms;
+        }
+
+        if (count_ms < 0) {
```
miyatakazuya: I don't think this timeout will ever trigger as on the 25th iteration `count_ms = 0`, while loop exists but the timeout is not catched. 
> On second thought, can we just check `init_status != std::future_status::ready` instead?

miyatakazuya (changes requested):
Big thanks for fixing this. I've noticed some extraneous comments + potential bug on the path generation timeout. I didn't SITL test it yet, but should be addressed. 

Also looking at the TODOs on the PR, I think the need to add a "path generation" tick is urgent. As is, if the path generation does time out and return nullptr, the `isMissionFinished()` may still be true because a new mission isn't uploaded yet, which would cause the lap count to decrement without the path generation succeeding.

src/pathing/static.cpp
AskewParity: The point that the plane will be in when it starts fly search is the last waypoint.

src/ticks/fly_waypoints.cpp
AskewParity: resolved

src/pathing/static.cpp
AskewParity: resolved

miyatakazuya (commented):
The change to MissionState is great. Much better way to modify the mission. 

I do think that we still have the flaw where if the path generation times out, the `tick()` function will just rerun since I think the `LOG_F(ERROR,...)` doesnt actually exit the code which breaks the numlaps logic if the MissionFinished is not updated properly which it likely is not since a mission did not get uplodaded. Again this should be addressed via a different tick, but a bandaid solution would to stop execution if the path generation finished OR have state var in the tick like `is_generating_path` which if true does not decrement the laps. 

Or we could just work on the new tick feature, assuming the path code will not time out for now lol. Fine with either, but @kimichenn should sitl test this b/c my ubuntu docker env is acting up and i need to fix it.

kimichenn:
I have already sitl tested it during the meeting today. Most of the things work fine except for one of the waypoints, which had a weird loop around it for the generated path. Chris said it's related to computational efficiency, which can be resolved. (Waypoint 3 below)

image

AskewParity:
> The change to MissionState is great. Much better way to modify the mission.
> 
> I do think that we still have the flaw where if the path generation times out, the `tick()` function will just rerun since I think the `LOG_F(ERROR,...)` doesnt actually exit the code which breaks the numlaps logic if the MissionFinished is not updated properly which it likely is not since a mission did not get uplodaded. Again this should be addressed via a different tick, but a bandaid solution would to stop execution if the path generation finished OR have state var in the tick like `is_generating_path` which if true does not decrement the laps.
> 
> Or we could just work on the new tick feature, assuming the path code will not time out for now lol. Fine with either, but @kimichenn should sitl test this b/c my ubuntu docker env is acting up and i need to fix it.

oops, I meant to put decrement path after it succesfully generated a new path.

miyatakazuya (approved):
LGTM 🦖 

> I lowkenuenly didn't test the last commit but it looks fine 👍I also think the circular paths adds personality (we should fix). 

Next steps are definitely to benchmark our pathing behavior in HITL testing

PR #348, opened 2026-02-27
chore: restore mav updates

AskewParity:
Soooooooo, apparently, our mav implementation doesn't impliment our own specificaiton

- add subscribe for heading
- restore subscribe for yaw, pitch, roll (I don't know why this was commented out in the first place)

Nothing has been tested...

### DIFF
--- a/src/network/mavlink.cpp
+++ b/src/network/mavlink.cpp
@@ -173,6 +173,13 @@ MavlinkClient::MavlinkClient(OBCConfig config)
         this->data.armed = armed;
     });
 
+    this->telemetry->subscribe_heading([this](mavsdk::Telemetry::Heading heading) {
+        VLOG_F(DEBUG, "Heading: %d", heading.heading_deg);
+        Lock lock(this->data_mut);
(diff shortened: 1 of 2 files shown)


miyatakazuya (commented):
The changes look good, but do we ever use IMU data in the obcpp controls? Same with heading.

AskewParity:
Yes, read the next PR (oops, I forgot to undraft it). I don't think we use IMU data, would you like me to delete it?

miyatakazuya:
@kimichenn Can you verify this PR works on SITL? Having some issues wit my ubuntu env atm. 
You can probably modify `tests/integration/mavlink_client.cpp` to output heading yaw pitch roll to test function.

kimichenn (approved):
lgtm. Everything looks okish from sitl, though the pitch is kinda sketch, but that can just be us being in quadcopter instead of quadplane. We could test it on Lek's quad, but mav is too nasty to work with. Let's just hope nothing breaks when we do fly.

PR #356, opened 2026-04-03
Added client side heartbeat ping

### LINKED ISSUE #354: PI-Client Heartbeat
Implement a heartbeat: (`CameraRequest::PING`) which sends a request to the server with some UID, and expects back the same UID if the server is responsive.
- This is for us to fall back on, as a method to check if the server is alive during setup + in requests prior to sending a image request.
- The UID would ensure that the response that is sent back matches the most recent ping.

CSTBook:
Closes #354.


miyatakazuya (changes requested):
The CI doesn't pass because of a mission semicolon in `udp_client.hpp`. Also the lint does not pass as well. 

Also is there an associated PR in camera-things? I'm a bit confused on the structure of the PR

include/network/udp_client.hpp
```diff
@@ -36,6 +36,7 @@ class UDPClient {
         Header recvHeader();
 
         std::vector<std::uint8_t> recvBody(const int mem_size, const int total_chunks);
+        char recvPing()
```
AskewParity: ;

miyatakazuya (approved):
lgtm 🦖

Issue #350, opened 2026-02-27
Dynamic Coverage Pathing


AskewParity:
Currently, the coverage pathing is being generated when all paths are generated. This leads to an inefficient path that may force ardupilot to loop around for waypoints that are otherwise colinear.

We can instead generate converage plans on demand, which will eliminate this issue.
This is contingent on changing the tick code, and making sure that pathing runs fast enough on the jetson.

PR #361, opened 2026-05-08
Feat/precompute pathing

### LINKED ISSUE #342: Lap Race Condition
## To Reproduce

To do this synthetically (SITl), we will have to change the code.

1. After the first completed waypoint mission, **don't** upload a new waypoint mission.
2. The code will transition to `fly_search`, which is incorrect as it only finished one lap.

## Expected behavior

The correct behavior is that the code should be in `fly_waypoints` until the number of laps finishes

## Additional context

There is a race condition between mav upload and the tick itself. If the tick reloads (on the order of 100 ms) before the mav upload completes, and the mav status is still marked as complete, then another lap will be subtracted.

AskewParity:
closes #342 #350 #360... sorry for the bloat

# Changes

- Add new item in `mission_state::next_waypoint_path` and associated functions, which is the waypoint-->waypoint path
- `path_gen` now precomputes all new paths (they must be in sequence now)
- `fly_waypoints` no longer generates new paths
- add new method in `static::calculateFinalAngle()` to rougly approximate the angle of an existing path
- unit tests for the `calculateFinalAngle`

# Testing

- There exist unit tests, you should probably think of the scope of the unit tests being larger (or to encompass other parts of the pr)

# TODO

- fix the remaining pathing unit tests
- naming
- angles. The current error in angle is pretty big (anywhere between 10-20 deg.)
  - It may be possible to retrieve the angle from the RRT process, but right now, that data is encapsulated


include/pathing/static.hpp
```diff
@@ -302,11 +302,26 @@ class AirdropApproachPathing {
...
-MissionPath generateInitialPath(std::shared_ptr<MissionState> state);
+/**
+ * Calculates the approximate angle of exit out of a waypoint path
+ * 
+ * @param path  ==> the waypoint path
+ * @param state ==> mission state
+ * @return          angle of exit (+x-axis is 0 deg, ccw is positive)
+ */
+double calculateFinalAngle(const MissionPath& path,
+                           const std::optional<CartesianConverter<GPSProtoVec>> cartesianConverter);
```
miyatakazuya: `cartesianConverter` can be a reference to avoid a copy

include/pathing/static.hpp
```diff
@@ -302,11 +302,26 @@ class AirdropApproachPathing {
...
+std::vector<GPSCoord>
+generateNextWaypointPath(std::shared_ptr<MissionState> state, double start_angle);
 
-MissionPath generateSearchPath(std::shared_ptr<MissionState> state);
+std::vector<GPSCoord>
+generateSearchPath(std::shared_ptr<MissionState> state, double start_angle);
 
-MissionPath generateAirdropApproach(std::shared_ptr<MissionState> state, const GPSCoord &goal);
+std::vector<GPSCoord>
+generateAirdropApproach(std::shared_ptr<MissionState> state, const GPSCoord &goal);
```
miyatakazuya, line 325: good change to make these return a `std::vector<GPSCoord>`. I think it makes the mutation of paths later on much easier if needed.

src/pathing/static.cpp
```diff
@@ -619,7 +619,23 @@ RRTPoint getCurrentLoc(std::shared_ptr<MissionState> state) {
...
+    const std::optional<CartesianConverter<GPSProtoVec>> cartesianConverter) {
+
+    const auto& coords = path.get();
+    if (coords.size() < 2) {
+        return 0.0;     // should probably either have angle between now and point
+                        // or use assert for force size >=2
+    }
+
+    XYZCoord pt1 = cartesianConverter->toXYZ(coords[coords.size() - 2]);
+    XYZCoord pt2 = cartesianConverter->toXYZ(coords[coords.size() - 1]);
```
miyatakazuya, line 627: Should `cartesianConverter` be optional at all? By the time `PathGenTick` is running the mission upload must have occured, so the converter should exist. 

I think its fine to keep it optional for our sanity, but then we should check the optional has a value before dereferencing it here

src/pathing/static.cpp
```diff
@@ -653,13 +669,53 @@ MissionPath generateInitialPath(std::shared_ptr<MissionState> state) {
...
+    RRT rrt(start, goals, SEARCH_RADIUS, state->mission_params.getFlightBoundary(), state->config,
+            {}, {});
+
+    rrt.run();
+
+    std::vector<XYZCoord> path = rrt.getPointsToGoal();
+
+    std::vector<GPSCoord> output_coords;
+    for (const XYZCoord &waypoint : path) {
+        output_coords.push_back(state->getCartesianConverter()->toLatLng(waypoint));
```
miyatakazuya, line 703: unsafe optional reference with `cartesianConverter` (i feel like i have something against cartesian converter or smth idk why i keep spotting these

src/pathing/static.cpp
```diff
@@ -653,13 +669,53 @@ MissionPath generateInitialPath(std::shared_ptr<MissionState> state) {
...
+    }
+
+    return output_coords;
 }
 
-MissionPath generateSearchPath(std::shared_ptr<MissionState> state) {
+std::vector<GPSCoord> generateSearchPath(std::shared_ptr<MissionState> state, double start_angle) {
```
miyatakazuya, line 709: missing a bounds check for the waypoints? I might be misinterpreting it but why is there a bounds check in `generateNextWaypointPath` to  check that we have at least 1 waypoint but not here?

src/ticks/fly_search.cpp
```diff
@@ -23,32 +23,17 @@ std::chrono::milliseconds FlySearchTick::getWait() const {
 }
 
 void FlySearchTick::init() {
+    // TODO: can we delete the following camera lines?
```
miyatakazuya: Vidhu's PR should remove these

src/ticks/fly_search.cpp
```diff
@@ -23,32 +23,17 @@ std::chrono::milliseconds FlySearchTick::getWait() const {
...
-
     LOG_F(INFO, "Total Waypoint #: %zu", this->state->getMav()->totalWaypoints());
 }
 
 Tick* FlySearchTick::tick() {
-    if (!this->mission_started) {
-        this->mission_started = this->state->getMav()->startMission();
-        return nullptr;
-    }
-
```
miyatakazuya, line 58: What is the rationale for removing this check? My understanding was that this protectes against the case that the `startMission()` command in the init of this tick is dropped. Without it, if the command is not successful then nothing will trigger the mission to start. I'm not in favor of the removal, but I also think the Mavlink startmision behavior has always been a bit confusing/off so it might be better to explore a different failsafe/state check.

src/ticks/path_gen.cpp
```diff
@@ -22,20 +22,41 @@ void PathGenTick::init() {
 }
 
 Tick* PathGenTick::tick() {
-    auto init_status = this->init_path.wait_for(0ms);
-    auto search_status = this->coverage_path.wait_for(0ms);
-    if (init_status == std::future_status::ready &&
-        search_status == std::future_status::ready) {
+    auto status = this->paths_future.wait_for(0ms);
+    if (status == std::future_status::ready) {
         LOG_F(INFO, "Initial and Coverage paths generated");
```
miyatakazuya, line 28: should update this log since we generate the next waypoint paths as well

src/ticks/path_gen.cpp
```diff
@@ -22,20 +22,41 @@ void PathGenTick::init() {
...
+
+        std::vector<GPSCoord> next_gps = generateNextWaypointPath(this->state, angle1);
+        MissionPath next = MissionPath(MissionPath::Type::FORWARD, next_gps);
+        double angle2 = calculateFinalAngle(next, this->state->getCartesianConverter());
+
+        int num_waypoints_to_remove =
+            std::ceil(this->state->config.pathing.upload_distance_buffer_m /
+                      this->state->config.pathing.dubins.point_separation);
+         std::vector<GPSCoord> coverage_gps = generateSearchPath(this->state, angle2);
+         coverage_gps.erase(coverage_gps.begin(), coverage_gps.begin() + num_waypoints_to_remove);;
```
miyatakazuya: this vector erasure is unsafe since `num_waypoints_to_remove` is not bound. also double semicolon

miyatakazuya (changes requested):
I left a few comments on some minor syntax issues and error handling, but overall the changes are good. The exit angle calculation although not nearly perfect does seem to be much more performant in terms nicer looking sitl paths. The PR also makes our encapsulation alot better in our ticks. I think once the comments are addressed I'll re-review really quick and it should be good to merge 🦖

src/ticks/path_gen.cpp
AskewParity: This log still conveys the same information.

src/pathing/static.cpp
AskewParity: The number of waypoints must be greater than 1 for the code to execute generate SearchPath.

miyatakazuya (approved):
The changes implemented addresses the comments well. I think `optional` for the cartesianConverter can be addressed at some point, but the `assert` solution is fine with me for now. LGTM 🦖

PR #362, opened 2026-05-08
Safely Terminate Aggregator

### LINKED ISSUE #324: Terminate Aggregator Early
Currently, we don't have a clean way of terminating CV detections. This is bad if we have too many images and can't process all of them in time. To fix this, we'll kill the aggregator process on submit our cluster results. This will be done in two parts:

Pt 1: Modify aggregator to kill all current processes cleanly, then move from the current tick (CVLoiter) to the next tick (AirdropPrep).

Pt 2: Modify the route associated with it to call this aggregator end process.

kimichenn:
Closes #324 

We now:
- Terminate aggregator safely and early when targets/matched is hit, and move to CVLoiter, then automatically to AirdropPrep
- Remove the check for whether we are in CVLoiter. If not, we move to CVLoiter then automatically to AirdropPrep

### DIFF
--- a/include/cv/aggregator.hpp
+++ b/include/cv/aggregator.hpp
@@ -3,15 +3,16 @@
 
 #include <atomic>
 #include <cmath>
+#include <condition_variable>
 #include <functional>
 #include <future>
+#include <map>
 #include <memory>
 #include <mutex>
 #include <queue>
 #include <thread>
 #include <unordered_map>
 #include <vector>
-#include <map>
 
 #include "cv/pipeline.hpp"
 #include "cv/utilities.hpp"
@@ -42,6 +43,9 @@ class CVAggregator {
     // Spawn a thread to run the pipeline on the given imageData
     void runPipeline(const ImageData& image);
 
+    // Stop accepting work, discard queued images, and wait for active workers to finish
+    void terminate();
+
     // Lockable pointer to retrieve aggregator results
     LockPtr<CVResults> getResults();
 
@@ -67,6 +71,8 @@ class CVAggregator {
 
     std::mutex mut;
     int num_worker_threads;
+    bool terminating;
+    std::condition_variable workers_done_cv;
 
     // For when too many pipelines are active at once
     std::queue<ImageData> overflow_queue;
--- a/include/cv/aggregator.hpp
+++ b/include/cv/aggregator.hpp
@@ -3,7 +3,6 @@
 
 #include <atomic>
 #include <cmath>
-#include <condition_variable>
 #include <functional>
 #include <future>
 #include <map>
@@ -70,9 +69,9 @@ class CVAggregator {
     Pipeline pipeline;
 
     std::mutex mut;
-    int num_worker_threads;
-    bool terminating;
-    std::condition_variable workers_done_cv;
+    std::atomic<int> num_worker_threads;
+    std::atomic<bool> accepting_images;
+    std::vector<std::thread> worker_threads;
 
     // For when too many pipelines are active at once
     std::queue<ImageData> overflow_queue;
--- a/src/cv/aggregator.cpp
+++ b/src/cv/aggregator.cpp
@@ -1,5 +1,7 @@
 #include "cv/aggregator.hpp"
 
+#include <exception>
+
 #include "utilities/constants.hpp"
 #include "utilities/lockptr.hpp"
 #include "utilities/locks.hpp"
@@ -7,6 +9,7 @@
 
 CVAggregator::CVAggregator(Pipeline&& p) : pipeline(std::move(p)) {
     this->num_worker_threads = 0;
+    this->terminating = false;
     this->results = std::make_shared<CVResults>();
     this->matched_results = std::make_shared<MatchedResults>();
     this->cv_record = std::make_shared<std::map<int, IdentifiedTarget>>();
@@ -26,7 +29,7 @@ CVAggregator::CVAggregator(Pipeline&& p) : pipeline(std::move(p)) {
     this->matched_results->matched_airdrop[AirdropType::Beacon] = dummy;
 }
 
-CVAggregator::~CVAggregator() {}
+CVAggregator::~CVAggregator() { this->terminate(); }
 
 LockPtr<CVResults> CVAggregator::getResults() {
     return LockPtr<CVResults>(this->results, &this->mut);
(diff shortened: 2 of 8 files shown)


src/cv/aggregator.cpp
```diff
@@ -63,8 +71,35 @@ void CVAggregator::runPipeline(const ImageData& image) {
...
+        std::thread worker_thread(&CVAggregator::worker, this, image, ++thread_counter);
+        worker_thread.detach();
+    } catch (const std::exception& err) {
+        --this->num_worker_threads;
+        if (this->num_worker_threads == 0) {
+            this->workers_done_cv.notify_all();
+        }
+        LOG_F(ERROR, "Failed to spawn CVAggregator worker: %s", err.what());
+    }
+}
```
AskewParity, line 81: 1. Why is this try-catched now? What is the failure mode? I don't necessarily think this is bad, but the change does deserve an explination.
2. The `num_worker_threads` should only be incremented after the thread starts instead of subtracting if the thread fails to start.

src/cv/aggregator.cpp
AskewParity, line 110: It seems `this->terminating` is checked multiple times for efficiency; however

1. It doesn't break out on new line 136;
2. I'm skeptical about the need for this efficiency.

src/cv/aggregator.cpp
```diff
@@ -63,8 +71,35 @@ void CVAggregator::runPipeline(const ImageData& image) {
 
     static int thread_counter = 0;
     ++this->num_worker_threads;
-    std::thread worker_thread(&CVAggregator::worker, this, image, ++thread_counter);
-    worker_thread.detach();  // We don’t need to join in the caller
```
AskewParity, line 67: Instead of detaching, we can keep track of the treads and join them later (in `terminate`).

src/cv/aggregator.cpp
```diff
@@ -63,8 +71,35 @@ void CVAggregator::runPipeline(const ImageData& image) {
...
+    try {
+        std::thread worker_thread(&CVAggregator::worker, this, image, ++thread_counter);
+        worker_thread.detach();
+    } catch (const std::exception& err) {
+        --this->num_worker_threads;
+        if (this->num_worker_threads == 0) {
+            this->workers_done_cv.notify_all();
+        }
```
AskewParity: The function call to `notify_all` will happen twice, it probably shouldn't.

include/cv/aggregator.hpp
```diff
@@ -67,6 +71,8 @@ class CVAggregator {
 
     std::mutex mut;
     int num_worker_threads;
+    bool terminating;
+    std::condition_variable workers_done_cv;
```
AskewParity: Is there a reason why we use this over c++ thread management?

src/network/gcs_routes.cpp
```diff
@@ -414,29 +414,35 @@ DEF_GCS_HANDLE(Post, targets, matched) {
...
     auto lock_ptr = state->getTickLockPtr<CVLoiterTick>();
-    if (!lock_ptr.has_value()) {
-        LOG_RESPONSE(WARNING, "Not currently in Loiter Tick", BAD_REQUEST);
-        return;
+    if (lock_ptr.has_value()) {
+        lock_ptr->data->setStatus(CVLoiterTick::Status::Validated);
+    } else {
+        CVLoiterTick* cv_loiter_tick = new CVLoiterTick(state);
+        cv_loiter_tick->setStatus(CVLoiterTick::Status::Validated);
+        state->setTick(cv_loiter_tick);
```
AskewParity: I am unsure if we ever switch ticks in this fashion - I'm uneasy about this.

n alternative is to have target validation be a part of mission state, and have `fly_search` also have a termination condition be the target validation. I think this would also make a bit more sense architecturally; although instead of a boolean it would be whether the coordinates are nullptrs or something.

Could @miyatakazuya weigh in on this?

AskewParity (changes requested):
Thanks for doing this, it should free a good amount of resources if our CV models are decent Lol.

I think this PR could be accomplished using C++'s thread management itself. That method probably reduces the complexity and increases the stability/reliability of shutting down the threads.

There were originally more comments over specific implementation techniques, but we can revisit them at a later time if we are keeping this solution.

src/cv/aggregator.cpp
miyatakazuya: I think `std::thread` can throw an error if its capped on resources so this prevents it from crashing if we hit compute limits.

src/network/gcs_routes.cpp
miyatakazuya: Yeah we shouldn't make a habit of calling setTick from external sources until we actually implement a genearl arbitary tick movement feature. It kind of violates the encapsulation of the state machine so I don't think its good practice to have it either. 

I agree we should probably have some `MissionState` var like `isTargetsValidated`, and here we set `state->getCV()->setTargetsValidated(true)` instead of overriding the tick directly.

src/network/gcs_routes.cpp
```diff
@@ -414,29 +414,35 @@ DEF_GCS_HANDLE(Post, targets, matched) {
...
+    state->getCV()->terminate();
 
-    if (matched_results.data == nullptr) {
-        LOG_S(ERROR) << "lockptr is null";
-    }
+    {
+        LockPtr<MatchedResults> matched_results = state->getCV()->getMatchedResults();
+
+        if (matched_results.data == nullptr) {
+            LOG_S(ERROR) << "lockptr is null";
```
miyatakazuya, line 421: i think this is an pre-existing bug but this shoudl return since if `matched_results.data` is null, then the latter code will just seg fault.

src/cv/aggregator.cpp
```diff
@@ -107,10 +145,9 @@ void CVAggregator::worker(ImageData image, int thread_num) {
 
     // 4) Mark ourselves as finished
     {
-        Lock lock(this->mut);
+        const int active_workers = this->num_worker_threads.fetch_sub(1);
         LOG_F(INFO, "CVAggregator worker #%d terminating. Active threads: %d -> %d", thread_num,
-              this->num_worker_threads, this->num_worker_threads - 1);
-        --this->num_worker_threads;
+              active_workers, active_workers - 1);
```
AskewParity, line 150: I think this math is wrong since you already subtracted one in line 148

AskewParity (changes requested):
Thanks for making this rewrite of the PR. I think it is more maintainable now.

I just have one clerical change to make.

src/cv/aggregator.cpp
kimichenn: `fetch_sub(1)` fetches, returns the value before subtraction, then subtracts, so this should be correct

https://www.tutorialspoint.com/cpp_standard_library/cpp_fetch_sub.htm

AskewParity (approved):

